PR #13, opened 2025-02-17
GitHub Actions

Hunter32R:
The neural network wrote this to me and it seems to work, the firmware is being compiled.


romasku:

At first glance, this looks like a great addition! The only issue I see is that it generates `.../refs/heads/main/...` links in the OTA index files. These links are not permanent, meaning the files they point to will eventually change. This causes problems when someone downloads an index file but later cannot upgrade because the links become outdated and the hashes no longer match.

To address this, I currently use a two-commit process: the first commit contains the updated build files, and the second replaces `.../refs/heads/main/...` with the commit hash, effectively making the link permanent. Could you please update your workflow file to use `make freeze_ota_links` to follow this approach?

.github/workflows/build.yml
```diff
@@ -0,0 +1,61 @@
...
+      run: |
+        make freeze_ota_links
+
+    - name: Commit changes
+      run: |
+        git config --global user.name "GitHub Actions"
+        git config --global user.email "actions@github.com"
+        git add bin/* zigbee2mqtt/*
+        git diff --cached --quiet || git commit -m "Update firmware files and converters"
+        git push
```
romasku, line 61: This unfortunately is not enough to freeze link propertly, a commit with bin files should be created first, then `make freeze_ota_links` then another commit. A something like this is needed:
```suggestion 
    - name: Commit changes
      run: |
        git config --global user.name "GitHub Actions"
        git config --global user.email "actions@github.com"
        git add bin/* zigbee2mqtt/*
        git diff --cached --quiet || git commit -m "Update firmware files and converters"
        make freeze_ota_links
        git add zigbee2mqtt/*
        git diff --cached --quiet || git commit -m "Freeze links in OTA index files"
        git push
```

PR #23, opened 2025-03-08
Add support for _TZ3000_f2slq5pj (TS0012) aka Bseed 2 Gang touch panel, no neutral

Oihso:
Switch is based on the ZTU module. Firmware is the same as Moes 2 Gang ED, but the pinout is different:
`Bseed-2-gang;Bseed-2-gang-ED;SB6u;SA1u;RD3;RC0;IC2;IB4;M;`

Already tested the firmware with the modified pinout - everything seems to work fine. Also there is a hardware possibility to fully disable both LEDs by using pin `C3`, but it needs to be implemented in software


romasku:
Hello,
Thank you very much for your contribution!

All files inside the `zigbee2mqtt` folder are auto-generated by the GitHub actions pipeline, so there was no need to update them.

PR #30, opened 2025-03-21
Add Moes MS-104CZ (Tuya TS0003) support

Hydron-4hv:
Added the device above - it's a 3-way switch module, neutral required, Aliexpress link: https://www.aliexpress.com/item/1005004986958450.html
Correct pic is here, Z2M just shows generic TUYA logo: https://user-images.githubusercontent.com/14648014/210105595-7cf007dd-ee0f-4cfd-82ca-79acf3268764.png
Note that this module has a buzzer on pin C2 (used on stock FW for indicating when it's in pairing mode) but it needs to be driven at ~3kHz to work and I didn't dig enough into your code to work out how to drive it. No indicator LEDs.

### DIFF
--- a/device_db.yaml
+++ b/device_db.yaml
@@ -370,3 +370,16 @@ TS0011_AVATTO_END_DEVICE:
   human_name: Avatto TS0011
   status: Supported
   github_issue: https://github.com/romasku/tuya-zigbee-switch/issues/16
+TS0003:
+  name: TS0003-custom
+  config_str: Tuya-TS0003-custom;TS0003-custom;BD3u;SC1u;SD7u;SC3u;RB5;RD4;RB4;
+  device_type: router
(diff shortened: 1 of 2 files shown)


romasku:
Many thanks for your contribution!

I'll open an issue for the buzzer, it should not be that hard to add.

Hydron-4hv:
Thanks to you for all your work on this software, it's allowing me to get much better use out of a bunch of switches I already had (I have some Moes TS0011/12s I've also got it running on), e.g. for detached mode and binding to a smart bulb (keeps the Mrs happy when my HA server is down, as the bound switches still work!)

I will test the buzzer support once you get it going. Attached is a scope shot of what the pin driving it was doing (it's about 3kHz, forgot to put the frequency measurement on sorry!).
Tuya_buzzer

PR #59, opened 2025-06-25
Release v1.0.16 - TOGGLE_SMART switch action modes

andrei-lazarov:
## Problem
I use bindings to control a group of smart bulbs with the physical button connected to the tuya switch.
The original TOGGLE behaviour individually toggles the bulbs, instead of toggling the group. It also ignores the relay state. 

## Fix
I renamed the old action to TOGGLE_SIMPLE and added 2 new modes to fix my problem:
- TOGGLE_SMART_SYNC: the bound devices will follow the relay state
- TOGGLE_SMART_OPPOSITE: the bound devices will always be in the opposite state of the relay

### TOGGLE_SIMPLE
TOGGLE_SIMPLE
### TOGGLE_SMART_SYNC
TOGGLE_SMART_SYNC
### TOGGLE_SMART_OPPOSITE
TOGGLE_SMART_OPPOSITE

## Zigbee2MQTT
Zigbee2MQTT

Building will generate new converters. So after the OTA installs, upload the new switch_custom.js, restart Z2M and perform interview and reconfiguration.


avbdr:
I have a feeling that the root cause is that toggle is having a bug that you are fixing with new modes introduction. 
Maybe it makes more sense to keep it simple and just update the standard toggle? @romasku

andrei-lazarov:
@avbdr I agree the drop down is really confusing if you don't have the context. We should document this better in the docs and maybe provide a link instead of the huge wall of text in the settings page.

Initially I also wanted to change the original toggle, but I think this enables new usecases people might use.  
If you connect an indicator LED to the relay output you can:
- use SYNC so the LED mimics the state of the lights. This way you know the state even if the lightbulb is not visible
- use OPPOSITE so the LED shines in the dark and you can find the switch easier

romasku:
Hi,

First of all, thank you for your contribution!

**TL;DR**: I like the idea of adding new toggle modes while keeping the old one for detached mode, to stay "mostly in line with the ZCL spec".

My initial idea was to follow [the ZCL specification](https://zigbeealliance.org/wp-content/uploads/2021/10/07-5123-08-Zigbee-Cluster-Library.pdf) in the hope of achieving better interoperability:

image

The idea in the specification is that your use case should be covered by the "ON_OFF" or "OFF_ON" options, which force synchronization between the physical state of the rocker switch and the relay/bound devices by ignoring some physical presses.

Personally, I also find it confusing when a press is ignored, so I use "toggle" mode. But here the spec says that the TOGGLE command should be issued, which effectively tells all bound devices to toggle, as you described. I want to keep this behavior, as it is useful in the detached state, where the relay is always on.

In a group context, TOGGLE indeed causes strange behavior, where individual devices toggle independently. Your new modes solve this issue when the relay is in use (and that is great!), but they don’t make sense in detached mode. Unfortunately, I don’t see a clear solution here, as the switch module is unaware of the state of the light bulbs. Even if we add some "memory" (in detached mode), if it sends "ON", then you turn off all bulbs from the app, and then do a physical press, it will send "OFF", effectively ignoring your press. 

I think only a centralized solution (via HA) can help in detached mode. But if you have a better suggestion, I’m open to discussion.

romasku:
I'm merging PR (code looks good, thank you), but we can continue the discussion.

andrei-lazarov:
Thanks for merging! I'm so happy I found a project I like that I can also contribute to 😄
  
> The idea in the specification is that your use case should be covered by the "ON_OFF" or "OFF_ON" 
  
It is, but only if used with a **classic switch**. I forgot to add, I am using **momentary switches** to avoid ignored presses.

I thought about this too:  
 
> but they don’t make sense in detached mode. Unfortunately, I don’t see a clear solution here, as the switch module is unaware of the state of the light bulbs
  
Indeed the 2 new modes don't make sense in detached mode, which is why we need one more mode to cover all usecases: **TOGGLE_SMART**.  

This mode should keep track of the bound lightgroup state **without using the internal relay.** So it could be used with either **attached or detached** relay mode.  

> Even if we add some "memory" (in detached mode), if it sends "ON", then you turn off all bulbs from the app, and then do a physical press, it will send "OFF", effectively ignoring your press.  
  
As you said, keeping the lightgroup state in a variable will not work because it will get out of sync when the group is controlled by another source. And the switch can't ask what state the group is in.

I have an idea though! 💡 Instead of a variable, we can create a fake / **virtual relay that we add in the lightgroup**.  
So a N-gang device will have N real relays and N fake relays. 
We could name them Helper relay 1, 2 .. N. Toggling them should not perform any action.  

When using the new mode **TOGGLE_SMART**:  
- Pressing the switch will toggle the real relay only if it is in attached mode.  
- Pressing the switch will always toggle the fake relay and send the corresponding ON/OFF command to bindings, based on the state of the fake relay.  
- The fake relay has to be added by the user in the lightgroup. So when the lightgroup is controlled by another source, the fake relay stays in sync.

Again this has the disadvantage of confusing the user with more buttons and endpoints. But it solves the problem and allows this usecase if anyone needs it.

Let me know what you think about this solution and how difficult it would be to implement.  
I'm not planning on using this, but if you like it and anyone needs it, I can experiment.

romasku:
I see your idea—you’re suggesting exposing this “hidden variable” as an OnOff cluster so that it can be controlled externally. Quite clever! It’s not completely foolproof, as someone might forget to add it to a group or control individual devices directly, but it should work.

However, it’s not that easy to implement and could be quite confusing. I think it can wait until someone runs into this issue. Then we can discuss what the best option would be for them.

PR #61, opened 2025-06-26
Update dependencies

andrei-lazarov:
The build workflow takes for granted unzip and wget.
Which is fair! They are preinstalled on the GitHub runner.

But I tried to build the firmware on WSL and it took me so long to figure out I was missing unzip!
If the packages were there I would not have run into that issue because I copy-pasted that line.

I think adding them makes it foolproof.


romasku:
Agreed, thank you.

PR #69, opened 2025-07-07
Update ZHA quirks and readme in workflow

andrei-lazarov:
I added `make_zha_quirk.py` and `make_readme.py` in the build workflow so the files stay up to date.  
Not sure why they were missing


romasku:
Thank you!

PR #72, opened 2025-07-10
Update relay.c

aceindy:
Bi-Stable Latch relays only require on/off pulse and should not be continuously powered. On BSeed 2 gang switches it leads to an endless boot loop, as the 2 coils draw 2x400mW continuously, which is too much without a neutral wire.

According specs the [HF3F-L/12-1HLT1](https://datasheet.octopart.com/HF3F-L-12-1HL1T%28257%29-Hongfa-datasheet-145013251.pdf) requires a pulse of at least 100ms

related to #67 

(sorry abt this new PR...messed up my repository)


andrei-lazarov:
I'm so glad you figured it out and managed to fix it. 😄 Good job!

PR #76, opened 2025-07-13
Avatto TS0003 - invert net led and shorter name

andrei-lazarov:
Similar to 2d67bc7a5ffc6fefeb0cdf1009f293171ca928e8, I updated Avatto TS0003:
- inverted network led
- shorter manufacturer and model name
- moved in-between Avatto 2 and 4 gang

Tested on my device and it works.


romasku:
Thank you!

Issue #70, opened 2025-07-09
Issue with BiStable relays (like in BSeed 2 gang drawing too much power)


aceindy:
@romasku I think the biStable relays should only be pulsed once, and not continuously powered which looks like is the case here:
`gpio_write(relay->off_pin, relay->on_high);`

Leaving it on would result in additional [Coil power consumption of 400mW](https://www.tme.eu/en/details/hf3f-l_12-1hl1t/miniature-electromagnetic-relays/hongfa-relay/), which in case of my dual channel would be 800mW 😉 

Related to https://github.com/romasku/tuya-zigbee-switch/pull/67#issuecomment-3051793197 / https://github.com/romasku/tuya-zigbee-switch/commit/10d591444150d27e3290aaf96f7d221a33bb85d1#commitcomment-161738183

aceindy:
This should solve it 👍 
https://github.com/romasku/tuya-zigbee-switch/pull/71

aceindy:
Hmm...I've tried several ways to make the code non-blocking with[ TL_ZB_TIMER_SCHEDULE](https://github.com/aceindy/tuya-zigbee-switch/blob/d58207ff2cd70720ef9cedd3a085e81c173c0351/src/base_components/relay.c), but every time the BSeed 2 gang becomes very unstable. Maybe a bit above my skills?
So maybe someone else can have a go at it ?
(For now I use my WaitMs(125), which works fine actually. I accept that when I push both switches simultaneously, one switch might not work...I just push it again 😄 )

romasku:
Hi, 
Got you. Can you please describe exactly how the device is unstable?

aceindy:
After a couple of times switching, the unit just seems to becomes un-responsive (both buttons as well as remote operation).
It requires a power reset to become operative again (the unit still reports to be online though).

rastakozza:
TL_ZB_TIMER_SCHEDULE documentation says:
Three uses of the return value of the timed task callback function.
- If the return value is less than 0, the task is automatically deleted after execution and the task ceases to exist.
- If the return value is equal to 0, the callback function always triggered periodically using the timeout in previous start-up.
- If the return value is greater than 0, the return value is used as the new period to trigger the callback function at regular intervals, in ms.

Did you try to add return -1 in schedule_pin_clear function?

andrei-lazarov:
This was fixed, so please close the issue 🙂

romasku:
[...]

PR #81, opened 2025-07-14
Latch relay125ms pulse

aceindy:
This time with a non-blocking scheduled pin clearing
Related to #70 and #72


aceindy:
Hmm... this PR is not stable...withdrawing it..

PR #82, opened 2025-07-14
Improve FAQ

andrei-lazarov:
I added more questions to FAQ. 
If anybody could answer the last one it would be great.

> Is it possible to revert to stock FW?


romasku:
Thank you, I'll add an answer to the questions about reverting to stock firmware.

PR #83, opened 2025-07-14
Fix tuya_man_id in Force image

andrei-lazarov:
I noticed the tuya_manufacturer_id is hardcoded in the Force image so I changed it. 
If it's supposed to be like that please reject this.


romasku:
No, your fix is correct, good catch, thank you.

Issue #85, opened 2025-07-15
New succesful porting _TZ3000_mhhxxjrs (3 gang switch)


gpaesano:
GREAT JOB!!!

<img width="576" height="524" alt="Image" src="https://github.com/user-attachments/assets/d976bc86-547b-4d4c-9a1d-9376f7889401" />

Just to inform on a succesful porting on
TUYA
TS0003
Zigbee Manufacturer
_TZ3000_mhhxxjrs (3 gang switch)

https://it.aliexpress.com/item/1005008107698143.html?spm=a2g0o.order_list.order_list_main.5.35853696iiCSQG&gatewayAdapt=glo2ita

No brand on the package. It is a three relay unit , based on TUYA ZTU module.

My config string: TS0003-3CH-cus;TS0003-3CH-cus;BC3u;LC2;SD7u;RD3;SB4u;RC0;SB5u;RD2;
Everythins works, led seems to be reversed (off when connection is ok).
Added in the device database and compiled locally, not really sucessful with OTA, applying config string in configuration via Z2M.
Not sure the 1, 2, 3 switch on the cover are correctly associated to 1,2,3 relay (but it is just a question of swapping wires), however it is not clear to me how to set the right order in the config.
In device_db.yaml:

TS0003_3CH:
  name: TS0003-3CH-cus
  config_str: TS0003-3CH-cus;TS0003-3CH-cus;BC3u;LC2;SD7u;RD3;SB4u;RC0;SB5u;RD2;
  device_type: router
  tuya_manufacturer_id: 4417
  tuya_image_type: 54179
  firmware_image_type: 43543
  stock_converter_model: TS0003_switch_module_2
  tuya_manufacturer_names:
    - _TZ3000_mhhxxjrs
  human_name: TS0003-3CH-cus
  status: WIP
  github_issue: https://github.com/romasku/tuya-zigbee-switch/issues/56

Also I have no clue about the firmware_image_type: is it a progressive number of the new firmwares?
The configuration should work also for the 2 gang and 4 gang (need to add RC1 and SB?).
Let me know if you need more info.

However flashed by wire everything went ok (uart speed at 921600) and used  this (copy of another 3 gang) converter:

{
        zigbeeModel: [
            "TS0003-3CH-cus",
        ],
        model: "TS0003-3CH-cus",
        vendor: "Tuya-custom",
        description: "Custom switch (https://github.com/romasku/tuya-zigbee-switch)",
        extend: [
[...]

PR #86, opened 2025-07-15
Add TS0003_3CH

andrei-lazarov:
Added the device from #85 as fw 43546


romasku:
Hi, thanks again for the PR!

Can you please fix the merge conflict? Also, it seems like `TS0003-3CH-cus;TS0003-3CH-cus;BC3u;LC2i;SD7u;RD2;SB4u;RD3;SB5u;RC0;` is the correct config (with correct order of relays as per @gpaesano)

andrei-lazarov:
@romasku Thanks! It should work now.

romasku:
Thank you!

Issue #77, opened 2025-07-13
Updating DEVICE CONFIG bricks unit


andrei-lazarov:
I updated my Avatto TS0003 with #76 to invert the network led.
I purposefully rebuilt the firmware and updated OTA instead of using the Device config field because @clumsy-stefan reported 2 bricked units after trying to change the pinout: #66 

After the update my device worked normally, but **I accidentally clicked on the text field**:

<img width="765" height="240" alt="Image" src="https://github.com/user-attachments/assets/1f187bcd-671c-427f-9724-53b6eff85ce6" />

**I did not change anything**, but the moment I clicked away from the text field, Z2M sent the command to update the config.
(There is no Submit button, so **Z2M automatically submits after every interaction**.)

I got a generic error and now my device is fully unresponsive:

```
z2m: Publish 'set' 'device_config' to 'AVATTO Triple Switch' failed: 
'Error: ZCL command 0xa4c1387ac2684c1e/1 
genBasic.write({"65280":{"value":
"TS0003-AV-CUS;TS0003-AV-CUS;BC2u;LD2i;SD3u;RC0;SD7u;RD4;SB6u;RC1;",
"type":68}}, {"timeout":30000,"disableResponse":false,"disableRecovery":false,
"disableDefaultResponse":true,"direction":0,"reservedBits":0,
"writeUndiv":false}) failed (Delivery failed for '47753'.)'
```

This makes 3 bricked units. We should investigate and maybe remove the option until we fix it? @romasku

clumsy-stefan:
Seems to be the same that happened to me...

Could you reflash it by wire?

The get function of the DeviceConfig did work for me though..

andrei-lazarov:
I do not have a UART flasher yet. But it would be useful to find the root cause of this issue so I might get one.

clumsy-stefan:
> This makes 3 bricked units. We should investigate and maybe remove the option until we fix it?

I guess just commenting out the `romasku.deviceConfig()` lines in `switch_custom.js` would be enough so that no one clicks accidentally in the field?

andrei-lazarov:
I got the UART flasher so I started debugging. I can probably figure this out

romasku:
@andrei-lazarov Just a note about debugging — adding `DEBUG=1` to the Makefile enables UART logs (via `DEBUG_INFO_TX_PIN`, which is set to `B1` by default). This makes debugging much easier.

andrei-lazarov:
[...]

PR #89, opened 2025-07-25
Urgent - Fix device config length 

andrei-lazarov:
Until now, the config string length (`config.size`) was never used.
We need it when reading and writing the device config string to and from NV flash memory to form the string correctly.

While I only targeted the length issue, this also fixes the bricking (as far as I can tell).
More info here #77.
I could not replicate the brick with these changes and the device is stable. (Tested on TS0003_AVATTO.)

However, I don't know how the 'memory slot' works.
How big is it? Is it ok that we write different-sized data into it?
Should we clear (remove) it before we write into it?

I was thinking to call the remove function in the write function, before writing. But I left it out currently as I was not sure.

Also, my Z2M instance fails to send strings longer than 74 chars (MESSAGE_TOO_LONG). So we should probably reduce the 256 limit. Is that based on anything?


romasku:
> However, I don't know how the 'memory slot' works.

To be honest, I also don't fully understand it either. Section 3.3.2 of [this document](https://wiki.telink-semi.cn/doc/an/AN_19052902-E_Telink%20802_15_4%20SDK%20User%20Guide.pdf) provides some information, but it's quite limited.

> So we should probably reduce the 256 limit. Is that based on anything?

I think I originally assumed it was the slot size limit, but I haven’t found anything that confirms that. I’m not sure we should reduce it here, but we should probably add some validation for `device_db.yaml`, since there were cases where the config for a newly added device was too large and was unreadable and unwritable in Z2M.

romasku:
> Is it ok that we write different-sized data into it?

It should be fine — the SDK itself uses it that way.

> Should we clear (remove) it before we write into it?

I don't think so. The firmware also uses it to store power-off behavior and other settings that need to persist across reboots, and it seems to work fine without clearing before writing.

My understanding is that this is some rudimentary filesystem-like structure designed to work around flash limitations, where data can only be erased (zeroed) in 4KB blocks.

andrei-lazarov:
Thanks for reviewing and answering! 
Indeed, I completely overlooked updating.
**I changed the slots as you suggested so they do not conflict between versions:**

- `NV_ITEM_ZCL_DEVICE_CONFIG_DATA = x + 1` (as it was)
- `NV_ITEM_ZCL_BASIC_CONFIG       = x + 2` (as it was)
- `NV_ITEM_ZCL_DEVICE_CONFIG_SIZE = x + 3` (new)

andrei-lazarov:
I did some more testing and I can confirm this update fixes the issue and does not cause any further damage..

- 1-4 gang devices that never had their config string interacted with - are unaffected.
- 3-4 gang switches that had their config string interacted with - got immediately bricked so they can't update.
- 1-2 gang switches that received a custom config string and survived - will update fine and revert to the default config string (while keeping the other settings). 
In this case I still suggest factory resetting before updating so the NV memory is cleaned.. (I tried to manually clear the old broken slot with lastSeenVersion, but any interaction with it bricks the device)

This is safe to merge and I suggest incrementing the version to push the update.

romasku:
Please resolve merge conflicts, and we'll finally publish this fix!

andrei-lazarov:
Thanks! The branch is up to date with main, I resolved the conflicts

romasku:
Hmm, it still shows as not resolved:

image

andrei-lazarov:
Screenshot From 2025-08-10 23-02-21

I guess something glitched out. I will create a new PR in a bit.

andrei-lazarov:
Moved to #116.

PR #90, opened 2025-07-28
Fix OTA for 4-Gang switches

clumsy-stefan:
When the config includes 4 switches and 4 relays OTA is not possible anymore due to not enough clusters (clusterId 25 is somehow overwritten and alwys reports "UNSUP_COMMAND").
Can be reproduced with any 3- or 4-Gang module, just enhance the config string.
Basically a size of 22 would be enough for 4-Gang switches, but to be ready for future enhancements I increased it to 32.


romasku:
Great catch, thank you!

PR #91, opened 2025-07-29
Fix Makefile

clumsy-stefan:
Enclose/escape DEFAULT_CONFIG with quotation marks as it contains semicolons that are interpreted by the shell, depending on the system/shell you use...


romasku:

PR #94, opened 2025-07-31
Always initialize EP1

clumsy-stefan:
Always initialize endpoint 1, even when no relays or switches are defined. 

This it will make it possible to create a "clean" universal binary with no config string in it and let the device still pair and connect to the ZB network. After that the unit can be configured in the UI.

Like this we could probably at some point simplify the various different images to a handfull different (if even needed). Also would. make it much easeir for OTA's.


romasku:
Thank you for your work!

PR #95, opened 2025-07-31
Add make_scripts for easier local builds

andrei-lazarov:
Hey all... local building was very confusing for me so here's my attempt as making it easier:

I created a new directory `make_scripts/` where you can find some scripts:

- `make_install.sh`: installs dependencies (**apt**) and (re)downloads sdk and toolchain.
- `make_all.sh` (already existed): builds fw for **all devices**; now it also updates converters, quirks and readme.
- `make_debug_single.sh`: build fw for **single device** with prints enabled, also converters, quirks and readme.

I use these on Ubuntu and I opted to get the dependencies from apt.  
Used pip and a virtual environment before, but I like this better.  
I will also suggest this approach on the GitHub Action workflow. Thoughts?

If you use other systems or different package managers please add more scripts or suggest changes.
Also, if this gets merged, I will update the documentation accordingly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@ PROJECT_NAME = tlc_switch
 BOARD ?= TS0012
 VERSION = 17
 
-DEBUG = 0
+DEBUG ?= 0
 
 SDK_DIR := sdk
 TOOLCHAIN_DIR := toolchain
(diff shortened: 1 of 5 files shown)


romasku:
Hi, looks great. I also like the idea of using these scripts in GitHub Actions to avoid duplication.

PR #96, opened 2025-07-31
Remove pip from build workflow

andrei-lazarov:
Cleaned-up the Python setup...  
We can get the dependencies from **apt** and avoid **pip**.

Tested on [remove_pip_build](https://github.com/andrei-lazarov/tuya-zigbee-switch/tree/remove_pip_build) branch and it works well.  
It generated the binaries, quirks, converters and readme.  

Here is the [test run](https://github.com/andrei-lazarov/tuya-zigbee-switch/actions/runs/16661783687) if you want to see the logs.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,17 +19,10 @@ jobs:
       with:
         fetch-depth: 0  # Получить полную историю коммитов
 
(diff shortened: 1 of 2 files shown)


romasku:
Thank you!

Issue #87, opened 2025-07-15
Same device, different pinout


andrei-lazarov:
I opened this issue to keep track of devices that have the same manufacturer ID, but different pin-outs.
(as we don't have a method to deal with this yet)

<details>
<summary>_TZ3000_5ajpkyq6</summary>

In #9 @paddenstoeltje reports:  
`BC2u;LD2i;SD7u;RB6;SC0u;RD4;SA0u;RC1;SA1u;RC4;`

In #9 @clumsy-stefan reports:  
`BC2u;LD2i;SD3u;RC0;SD7u;RD4;SB6u;RC1;SA0u;RC4;`

</details>

solevi:
Hi, back from vacation.
I apologize, I tested the device on my desk and somehow didn't notice that the ON/OFF logic was inverted (compared to the blue status LEDs state).
Indeed, @aceindy 's configuration is correct:
`Bseed-2-gang;Bseed-2-gang-ED;SB5u;SD4u;RC0B6;RA1D7;ID2;ID3;LC3;M;`
However, I am powering the device from a lab power supply with an ammeter, and I cannot confirm increased current draw with my configuration.
With both configurations, the current draw from the 12V power supply was around 84.1-84.5 mA (depending on LED states), pulsing during the ON/OFF state change.
Not sure what can cause instability.

andrei-lazarov:
@solevi Thanks for clearing it up! That's a relief 😅 I'll remove it from this list.
Do you need help getting the inverted configuration?
You should use the force index and perform an OTA.
Do not use the config string field. It might brick the device: #77

aceindy:
> With both configurations, the current draw from the 12V power supply was around 84.1-84.5 mA (depending on LED states), pulsing during the ON/OFF state change.
> Not sure what can cause instability.

@solevi  Which 'both' configurations?
With the current main repositories, the bistable relays are always powered, so there shouldn't be much difference between on/off (or pulsed).

You could re-test with https://github.com/aceindy/tuya-zigbee-switch/tree/LatchRelay125msPulse
That one does pulse on/off (125ms).
Since it uses a blocking code, it is not suitable for release, but it would allow you to measure if there is any difference in amps 🤔

solevi:
Hi, sorry, it seems I misread what you wrote. I thought that the increased current draw was due to my configuration option.
[...]

PR #99, opened 2025-08-01
Update Avatto blue switches

andrei-lazarov:
I made some updates for the AVATTO blue-socket switches.

Small changes: 
- shorter names (4-gang was still too long)
- old names to support devices on old fw versions
- correct Z2M names (will also add the correct picture)
- correct human names

Bigger change for the 4-gang switch (_TZ3000_5ajpkyq6):
- **correct pinout**

Me and @clumsy-stefan  recently got the switch from [AliExpress](https://vi.aliexpress.com/item/1005007247647375.html)  
and we report a different pinout than @paddenstoeltje, see #87.

The new pinout correctly follows the pattern from the 1-3 gang switches. 
And this is what you get when you buy the switch from the official international store, which should take priority over other sources.

Also, devices already flashed and working with the old pinout **can not update OTA** (#90) so they will never get the new pinout.

If someone reports the different pinout again, I will make a custom build for them.

### DIFF
--- a/device_db.yaml
+++ b/device_db.yaml
@@ -64,7 +64,7 @@ TS0002_OXT:
   tuya_manufacturer_names:
     - _TZ3000_bvrlqyj7
   human_name: OXT 2-gang switch
-  status: WIP
+  status: Supported
   github_issue: https://github.com/romasku/tuya-zigbee-switch/issues/49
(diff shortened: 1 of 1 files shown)


romasku:
Thank you!

PR #100, opened 2025-08-02
Fix/increase bind table size to 32

clumsy-stefan:
When bind table reaches 16 binds obviously no more binds can be made. But also the reportings can't be changed anymore as it always throws a "TABLE_FULL" error. If you bind all 4 buttons to some enddevice with onOff and levelCtrl (see PR #93) together with the standard bindings you get exactly 16 bindings, which make it impossible to change /add reporting! Size increased to 32.

### DIFF
--- a/src/configs/stack_cfg.h
+++ b/src/configs/stack_cfg.h
@@ -72,7 +72,7 @@
 /**
  *  @brief  APS: MAX number of binding table size
  */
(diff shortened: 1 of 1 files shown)


romasku:
Thank you! We should probably start tracking RAM usage soon, but I think there should be enough space for now.

PR #101, opened 2025-08-04
Fix `nv_flashSingleItemRemove()` to use correct parameters

clumsy-stefan:
The api specifies the call as:
`nv_sts_t nv_flashSingleItemRemove(u8 id, u8 itemId, u16 len);`
This PR corrects the use of the call.

### DIFF
--- a/src/device_config/config_nv.c
+++ b/src/device_config/config_nv.c
@@ -28,7 +28,7 @@ void device_config_write_to_nv()
 
 void device_config_remove_from_nv()
 {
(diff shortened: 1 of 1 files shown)


andrei-lazarov:
Great catch, Stefan!
I see the remove function is called only when the config string fails validation, so the fix can not cause any issues/instability.
 
Note that this fix is temporary, until I finish testing #89.
I will also update the remove there, to delete the same number of bytes as there are written (correct size as opposed to 256 bytes on main branch)

Thanks again for noticing the remove is wrong 😉

romasku:
Great catch!

Issue #105, opened 2025-08-06
New 4-Gang Avatto Clone Device _TZ3000_knoj8lpk


clumsy-stefan:
Found anothoer 2-Gang Avatto clone!

<img width="1518" height="1382" alt="Image" src="https://github.com/user-attachments/assets/e736e448-1cc1-409d-8c63-9b3b3077d436" />

<img width="1246" height="1247" alt="Image" src="https://github.com/user-attachments/assets/93022098-6ef6-4645-8014-896e9886c6ae" />

<img width="1366" height="1404" alt="Image" src="https://github.com/user-attachments/assets/0ff65182-7da6-4030-bd0b-5e621a96de45" />

Pinout / Config as follows:

BC3u;LC2i;SB5u;RD2;SB4u;RD3;SD7u;RC0;SD4u;RC1;

Generally, why do we change the manufacturer name and not keep the _TZ3000.. or whatever? Just chaning the Model or even image type would be enough to note, that it's a custom firmware...
Eg:

Zigbee Model: Custom Firmware
Zigbee Manufacturer: knoj8lpk (original second part of MF)
Image-Type: 43548 (uniqe curtom FW identifier)

Like this we would not loose the reference to the original device for future integrations..

And you could simplify quite good, one converter for the each of the 1-2-3-4 gang (probably 2 variants, depending if there are different needs, like bistable relays, etc.). Converter is selcted by the Zigbee Model (1-2-3-4 Gang etc.) and updates are selected by the (shortended) original manufacturer and image type...

Again, IMHO, to be commented if wanted...

romasku:
> Zigbee Model: Custom Firmware
Zigbee Manufacturer: knoj8lpk (original second part of MF)
Image-Type: 43548 (uniqe curtom FW identifier)

I like this idea. When the firmware was first developed, there weren't that many devices, and it was relatively easy to select meaningful names by hand. But as the list grew, it became somewhat messy. So I completely agree that having a systematic approach that also traces back to the original device is a good idea.

If you don't mind, please create a separate issue to implement this idea.

clumsy-stefan:
I opened new issue #110 for the information change/standardisation.

And as @andrei-lazarov made a PR that's already merged, I close this issue.

andrei-lazarov:
Added 1-gang variant (_TZ3000_pgq7ormg) as **TS0001_IHS**.
Pinout: `BC3u;LC2i;SB5u;RD2;`
[...]

PR #107, opened 2025-08-06
Add TS0004_IHS (_TZ3000_knoj8lpk)

andrei-lazarov:
Added the device from #105 and renamed its 3-gang brother


romasku:
Thank you for your work!

PR #108, opened 2025-08-06
Multiple pinouts warning

andrei-lazarov:
- added [multiple_pinouts doc](https://github.com/andrei-lazarov/tuya-zigbee-switch/blob/dfb6dcd45f1d583f613244433d77b156a09841f9/docs/multiple_pinouts.md)
- added warning in tuya converter
- added alternative config string in db for avb 4gang

There's one device with different pinout reports... so I wanted to give users a warning before they update.
My idea was to update the description with a big warning and a link to the docs. (see screenshot below)

Even if it's only one device, I took the time to automate it:

I suggest the new `alt_config_str` field in the database, which should keep alternative config strings/pinouts.
If it is populated, the script will automatically add the warning to the tuya converters.

Note that I already run the script so you can see the new converter.

If you have suggestions on how to make it more visible, please share.
I was also thinking about the patch notes message you get when you check something for updates. (IKEA bulbs for example)

Screenshot From 2025-08-06 23-53-18

### DIFF
--- a/device_db.yaml
+++ b/device_db.yaml
@@ -183,6 +183,8 @@ TS0004_AVATTO:
   tuya_image_type: 54179
   firmware_image_type: 43529
   stock_converter_model: TS0004_switch_module_2
+  alt_config_str:
+    - TS0004-AVB;TS0004-AVB;BC2u;LD2i;SD7u;RB6;SC0u;RD4;SA0u;RC1;SA1u;RC4;
   tuya_manufacturer_names:
     - _TZ3000_5ajpkyq6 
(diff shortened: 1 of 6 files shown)


clumsy-stefan:
Great idea!!

However, are you sure that the tuya manufacturer is really the same and the pinouts are different? We need to make sure it's not a accidental mismatch.. I never came across these in my switches, even though some of them I ordered 2 years ago and reordered now.

But as you already implemented, I think we should integrate it to be future proof...

romasku:
Sorry for the late review, thank you for your contribution!

Issue #106, opened 2025-08-06
Leave and rejoin ZB Network also clears NVRam config


clumsy-stefan:
When a device is removed from the network (either from the server side eg z2m leave or via config-button) all configuration stored in NVRam is cleared. Specifically the Button/Relay types, long press duration, etc. SO basically a factory reset is done.

I cam across this while trying to implement different reset options (one for multi-press switch, one for long press config button, one for multi-press config button). However no matter which reset method is used, as soon as the node leaves the ZB network all information is cleared. and a factory reset is done.

A similar method (different reset methods) is shown in [this](https://github.com/pvvx/ZigbeeTLc) project. However these various reset commands seem to behave differently within this firmware and always factory reset.

In the end the idea is to have a reboot, a reset network information and start binding and a factory reset option. Especially a "leave and rejoin" reset would be nice to be able to move the node between ZB networks without loosing switch/relay configs.

Any help appreciated!

Regards

STefan

andrei-lazarov:
We previously discussed this on #98 and everybody agrees it will be useful.  
I also confirmed this is how it works on original firmware.

Thanks for being the one to try it implementing it!
Can you please share what you already tried and didn't work?

We can see on the repo you linked: they were able to achieve 'leave without reset' with this function: [tl_bdbReset2FN](https://github.com/pvvx/ZigbeeTLc/blob/master/src/app_ui.c).
I understand you already tried that and it still clears NV memory. I'm curious why..

clumsy-stefan:
>Can you please share what you already tried and didn't work?

Tried a number of things, but didn't want to bore anyone here ;) (or probably didn't want to expose my cluelessness :joy:)

Differnent reset methods (one at a time obviously):
- `tl_bdbReset2FN();`
- `tl_bdbReset();`
- `zb_reset();`
- `zb_resetDevice();`
- `nv_resetAll();`
- `nv_resetModule(NV_MODULE_ZCL);` then reboot
- `zb_deviceFactoryNewSet(1);` then reset (or set to 0)
- `zb_assocJoinReq();`
- `SYSTEM_RESET();`
- `bdb_networkSteerStart();`

Or actively leaving the network:
```
     nlme_leave_req_t req;
     ZB_IEEE_ADDR_COPY(req.deviceAddr, g_zbMacPib.extAddress);
     req.rejoin = 0;
     req.removeChildren = 0;
     if(!g_bdbCtx.leaveDoing){
[...]

PR #109, opened 2025-08-07
Add lastSeenVersion

andrei-lazarov:
I suggest we write the firmware version in NV memory.

This will help us in the future to treat edge cases.
(Any change to the memory map that would cause incompatibilities between versions)

For example this will give us the freedom to change memory slots:
```
if (lastSeenVersion < 18 and currentVersion >= 18) {
    copy from old_slot to new_slot
    delete old_slot
}
```

However I can't get it to build the new file... I keep getting undefined reference to the new function.
Where do I need to define/declare it?

(do not merge yet)


clumsy-stefan:
I think that would be a nice thing, however I would not take firmware version (as this is already in the device, you can see it when you look at the OTA page in z2m for example) and you should be able to extract it. I'd rather take the latest commit hash to uniquely identify what's running on the node even when version number is not bumped. Also usefull for testing..

Sorry, I misunderstood the request I think... You wan't to keep track which config version is currently in ram vs. the one in the firmware... of course, that would be a great thing!!

Still, theoretically you could also use the commit hash for it, just makes the if condition a bit different..

I can have a look on where it needs to be defined. Have you already pushed a commit? Or send me the changes offline...

romasku:
>However I can't get it to build the new file... I keep getting an undefined reference to the new function.
Where do I need to define/declare it? 

All sources are manually listed in makefiles, and project files are listed in [this file](https://github.com/romasku/tuya-zigbee-switch/blob/main/makefiles/src.mk).

For me, your idea is very similar to DB migrations, which allow updating data to a new schema in newer versions. I like the idea!

clumsy-stefan:
BTW: when changing these config things, should we also look on where it stores?

Currently the information is stored in `NV_MODULE_ZCL`, from the documentation it says that user/application-data should be stored in `NV_MODULE_APP`. Not sure if this changes something, but just wanted it to mention.

I came across this while looking at the factory reset on unbind issue (#106), however at least my tries didn't change anything for that issue if it's stored in ZCL or APP partition, it still got deleted!

andrei-lazarov:
@clumsy-stefan 

> I'd rather take the latest commit hash to uniquely identify what's running on the node even when version number is not bumped. Also usefull for testing..

I'm open to adding an in-between versions identifier. I'm not sure the commit id is the way to go..  
(When developing and testing, you might flash by wire without committing).  

But we could work with that. Let's construct a string:  
`branch + version.small_version + custom_text + commit`  
and store it in an attribute.. so you could read it from the Z2M dev console.

On the main branch it would be always incremented by the build script:
`main 17.1 release #commit`  
While testing, you would manually edit a parameter (in your custom make script) for the custom_text:  
`levelCtrl 17.1 attempt_2_for_brightness_control #commit`
I ended up liking this idea. 😃

Thanks for checking the memory slots.. I will also try `NV_MODULE_APP`.

@romasku 
Thanks for your quick reply and approval of the idea!
I added the new file to the list. I built and tested it a bit (update OTA, update by wire, reboot, reset).  
It works as expected so we can merge this.

There is no need to increment to version 18 and push an update just for this.

Screenshot From 2025-08-07 19-39-43

romasku:
I think we should eventually move to `NV_MODULE_APP`, as it indeed sounds more appropriate for what we are doing, but let's do it in another PR, maybe using the mechanism introduced here.  
So, I'm merging this — thank you very much for implementing it.

PR #111, opened 2025-08-07
Update to newest Telink SDK V3.7.4.1 **Not for merging yet!!*

clumsy-stefan:
**NOT FOR MERGING YET!!**

*Be careful, this PR could eat your dog or even brick your device!!*

Before digging deeper into issues documented in #106 I thought it could be wise to update to the latest SDK as the current (V3.6.8.6) is over a year old. There were a number of low-level changes and as I think some performance/size improvments in the SDK...

[Version history here](https://github.com/telink-semi/telink_zigbee_sdk/releases)

As I was not clear why/what the `patch_sdk` fils/changes are usefull for I decided to stick to the original SDK versions and discarded these local adaptions. If someone knows why not the original SDK files were taken and what changes were made, please enlighten me!!

Building runs fine, also all the basic tests I did worked well. Switching, binding, pairing, OTA, etc. all work OK.

However I can't confirm 100% that it runs stable over a longer period, so I suggest to hold back merging this change until others also tested it! Having said that I'd appreciate if others could try and test!! Any feedback greatly appreciated!!


romasku:
The main problem was with OTA from the stock firmware: I remember trying to use a more recent SDK, but it caused the device to be bricked if flashed via OTA from the stock firmware. 

Regarding patches: If I remember correctly, the patches mainly mark `_attribute_ram_code_sec_` some parts of the SDK used by `tuya_zigbee_ota`, but I'm not sure.

clumsy-stefan:
Flashing custom firmware devices with the new SDK works without issues and seems to run stable (I think more stable and with improved performance).

Flashing a original firmware device seems to brick it, unfortunately I only had one original left to test... so I can't reproduce currently..

So I'm unsure if I should proceed on this or not. Probably it would be a possibility to first flash a old SDK custom FW and then update to the new SDK FW...

EDIT: at least you can switch back and forth between SDK's without issues on custom FW...

romasku:
Closing this, as `main` is now also using 3.7.2.0. Thank you very much for maintaining this version until `main` was able to catch up.

PR #112, opened 2025-08-10
Added update_converters.js

andrei-lazarov:
Hey all! I was thinking of a way to automatically update the custom converters..  
As not all users are aware you need to update them to support new devices and new features in the firmware.

I noticed that Z2M allows running any javascript code in the external_converters..  
**So I made a script (an optional converter) that updates the custom converters.**

If you put it in external_converters, **it gets executed every time Z2M starts**.  
Which is not ideal.. because you might never restart Z2M, but I believe this is a step in the right direction.  
(Please come with suggestions in this regard)

Also it is a bit weird.. because it downloads new converters while (or after?) loading the old converters.  
It does not look like they conflict, but the new ones are effective only after a restart.. which is why I added the restart MQTT command at the end of the script (needs MQTT credentials). 

It compares the md5 hash to check for updates. And it creates a backup of the local files if an update is detected.

I tested it on my setup and it works.. I did not run into errors, I could not get it to bootloop. If there is no internet connection it does not do anything.  
You might want to check if the external_converters path is the same on your systems.

What are your thoughts?  
I think the script is pretty invasive...  
It needs internet connection, access to the file-system, MQTT credentials, it restarts Z2M. It downloads and runs new scripts after all..
Surely Z2M does not have any protections against malicious code. This must come with a big warning presented to the user.
Though the same argument could be made to manually updating the converters (who checks their contents?).. or running any piece of software on your computer 😃 especially closed-source.

Realistically I think the script is safe.. The only risks are Romasku turning evil 😄 or his account getting hijacked.
At least I will use it, and I hope you will find it useful too.. anyway sorry for the huge wall of text.

### DIFF
--- a/zigbee2mqtt/converters/update_converters.js
+++ b/zigbee2mqtt/converters/update_converters.js
@@ -1,0 +1,123 @@
+const https = require('https');
+const fs = require('fs');
+const path = require('path');
+const crypto = require('crypto');
+const mqtt = require('mqtt');
+
+// Automatically updates the custom converters
+// for romasku firmware whenever Z2M starts.
+// It creates 1 backup of the files before updating.
+
+// Check the settings below and place it in
+// zigbee2mqtt/data/external_converters directory
+
+// WARNING!
+// This script is very invasive!
+// Use it at your own risk!
+// (like any other scripts you get from the internet)
+
+
+/************************** Settings ******************************/
+
+// URL
+const user = 'romasku';             // default: romasku
+const repo = 'tuya-zigbee-switch';  // default: tuya-zigbee-switch
+const branch = 'main';              // default: main
+
+const url_tuya_with_ota = `https://raw.githubusercontent.com/${user}/${repo}/refs/heads/${branch}/zigbee2mqtt/converters/tuya_with_ota.js`;
+const url_switch_custom = `https://raw.githubusercontent.com/${user}/${repo}/refs/heads/${branch}/zigbee2mqtt/converters/switch_custom.js`;
+
+// Local path to the converters directory
+const converters_path = 'data/external_converters';
+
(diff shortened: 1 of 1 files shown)


marazmarci:
Hi!

> I was thinking of a way to automatically update the custom converters..
> As not all users are aware you need to update them to support new devices and new features in the firmware.

I was also thinking of the same when installing the custom converter for the first time 😄

> I think the script is pretty invasive...
> It needs internet connection, access to the file-system, MQTT credentials, it restarts Z2M. It downloads and runs new scripts after all..

I agree... I think as a user it would be surprising if "random scripts" would change "random files" in my Z2M installation.

Did you think about upstreaming the custom converter into [zigbee-herdsman-converters](https://github.com/Koenkk/zigbee-herdsman-converters)? This way, the converter would only change when it is expected: when the user updates Z2M. Of course this would make the life of this project a bit harder, because we could only release new converter versions slowly, only together with Z2M, but I think it would be a cleaner approach. We could always tell users to go switch to the [Z2M dev branch](https://www.zigbee2mqtt.io/advanced/more/switch-to-dev-branch.html) if they want to test out the newest converter version.

andrei-lazarov:
@marazmarci thanks for your input!
I would love to see the converters on official Z2M 😃 That definitely is the goal.
But I thought it is a far-fetched idea so I did not even mention it.

- `switch_custom.js` will not conflict with anything so it's ok.
- `tuya_with_ota.js` will enable OTA for many devices on stock FW.. so I'm not sure if we can send that one..

Users still need to apply the custom index though.. 
so they can't accidentally update to custom FW (not that they would complain 😉)

I am currently trying to standardize the converters so we can lower their count and reuse them between devices.
Maybe after that I'll look up contributing to the official converters.

PR #114, opened 2025-08-10
Introduce unit tests support using Unity

romasku:
In this PR, I’ve added a way to write unit tests for the firmware.  
As the firmware grows larger, I believe we should start covering it with tests to ensure it remains stable.  
This also highlights the need for a HAL (hardware abstraction layer), which could be useful for porting to non-Telink devices in the future.  

Some details can be found in the new [documentation](https://github.com/romasku/tuya-zigbee-switch/blob/rs/unit-tests/docs/tests.md).

@andrei-lazarov @clumsy-stefan — if you have time, please review. I’d greatly appreciate your feedback.

### DIFF
--- a/docs/tests.md
+++ b/docs/tests.md
@@ -1,39 +1,26 @@
-# Unit tests
-
-## Executing tests
-
-```
-make tests
-```
-
-## Writing tests
-
-This project uses [Unity](https://github.com/ThrowTheSwitch/Unity) for unit testing. All test leave inside `tests/` subdirectory. Test suites use the following pattern:
-
-- Each test suite lives inside subdirectory under `tests/`, for example `tests/button/`. 
-- Test suite defin# Unit Tests
+# Unit Tests
 
 ## Executing Tests
 
 ```bash
 make tests
-````
+```
 
 ## Writing Tests
 
 This project uses [Unity](https://github.com/ThrowTheSwitch/Unity) for unit testing.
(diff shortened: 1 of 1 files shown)


clumsy-stefan:
I think that's an excellent idea... However I have absolutely no experience with automated testing environments...

I also like the idea about a HAL... not sure though how it could be done. I just saw that ESP's do have such a layer I think..

andrei-lazarov:
I did unit tests at my last workplace and I'm not thrilled to see them again 😅.
The general impression there.. was that we spent more time writing/updating/running tests instead of actual developing.

I can see the desire for some automated testing though.. It makes sure nothing breaks whenever we make changes to the source code.
However, I can't really imagine a case where they would be useful and catch some errors.
In my mind, the tests are either so simple they never fail (I don't think any addition will break button presses)  
or so specific they will fail on every change and will always need updating..
Usually the issues are sporadic, unpredictable and not identifiable in tests because you don't know what will go wrong when you write the tests.
But maybe I'm totally wrong and this is not the intended purpose.. or they turn out to be useful.

My opinions aside.. testing is a great sign of the projects' maturity and your commitment to it.  
I appreciate your effort for adding the testing framework! I can work with it! 
It looks good, it runs fast, it's easy to write and understand. There's no nonsense.

Please remove the duplicate paragraphs in the doc and we can merge this. Then we can slowly add more test coverage.

PR #115, opened 2025-08-10
Add support for genLevelCtrl cluster

romasku:
This is heavily based on https://github.com/romasku/tuya-zigbee-switch/pull/93 — thank you once again.
I haven’t tested it on a real device yet, so it’s still a draft for now, but the idea is already there. @clumsy-stefan, feel free to take a look if you want.

### DIFF
--- a/helper_scripts/templates/switch_custom.js.jinja
+++ b/helper_scripts/templates/switch_custom.js.jinja
@@ -41,10 +41,10 @@ const romasku = {
         enumLookup({
             name,
             endpointName,
-            lookup: { detached: 0, raise: 1, short_press: 3, long_press: 2},
+            lookup: { detached: 0, press_start: 1, short_press: 3, long_press: 2},
             cluster: "genOnOffSwitchCfg",
             attribute: { ID: 0xff01, type: 0x30 }, // Enum8
-            description: "When to trigger internal relay",
+            description: "When to turn on/off internal relay",
         }),
     relayIndex: (name, endpointName, relay_cnt) =>
         enumLookup({
@@ -61,10 +61,10 @@ const romasku = {
         enumLookup({
             name,
             endpointName,
-            lookup: { raise: 1, short_press: 3, long_press: 2},
+            lookup: { press_start: 1, short_press: 3, long_press: 2},
             cluster: "genOnOffSwitchCfg",
             attribute: { ID: 0xff05, type: 0x30 }, // Enum8
-            description: "When to trigger binded device",
+            description: "When turn on/off binded device",
         }),
     longPressDuration: (name, endpointName) =>
         numeric({


clumsy-stefan:
I'm happy to try it (not anymore today though ;) but would it be possible to very briefly explain what `binded mode` is meant for or what's the logic behind? Purely fom reading the code I think I don't yet understand.... thx..

clumsy-stefan:
Tried quickly today, I noticed the following points:

image

If I wan't to act the remote device as well on short press (onOff) as on longPress (levelCtrl), what should I select here?

 If I try to interact with the `bindedDevice` config I get an error:
 `z2m: Publish 'get' 'switch_0_binded_mode' to '0xa4c1381fd77320f5' failed: 'Error: ZCL command 0xa4c1381fd77320f5/1 genOnOffSwitchCfg.read([65285], {"timeout":10000,"disableResponse":false,"disableRecovery":false,"disableDefaultResponse":true,"direction":0,"reservedBits":0,"writeUndiv":false}) failed (Status 'UNSUPPORTED_ATTRIBUTE')'
[2025-08-11 08:00:00] debug:    z2m: Error: ZCL command 0xa4c1381fd77320f5/1 genOnOffSwitchCfg.read([65285], {"timeout":10000,"disableResponse":false,"disableRecovery":false,"disableDefaultResponse":true,"direction":0,"reservedBits":0,"writeUndiv":false}) failed (Status 'UNSUPPORTED_ATTRIBUTE')`

Unfortunately with this error I can't test further....

clumsy-stefan:
I think I found the problem...

in `switch_cluster.c` it should be:
`  info_conf->attrNum             = 8;`

as you introduced a new attribute...

clumsy-stefan:
I quickly tested with a real switch and bulp.

Unfortunately I have to report, that it doesn't work at all:
- short pressing doesn't seem to send the onOff command to the bound cluster
- long press doesn't work at all (at least not with factory settings).

Now as I said in the first note, I probably just don't understand how this "bind mode" works and you need to educate me. But out of the box doesn't do the job and the combinations I tried with the settings didn't work.

So my basic question: for a simple "short press local relay toggle and bound onOff clusters toggle" and "long press level control to bound LevelCtrl clusters", which combination should work?

I hope we'll get this fixed soon, as I think a number of people would love to see that functionality ;)

romasku:
I should be able to test it properly myself this evening — sorry about that.
I see your confusion about the term *"binded mode"*. This applies only to On/Off commands; level commands occur only during a long press.
I’ll definitely update the description to *"when to **turn on/turn off** the bound device"*, but that might still not be clear enough. Do you have any suggestions for a better name for the parameter itself?

> So my basic question: for a simple "short press local relay toggle and bound onOff clusters toggle" and "long press level control to bound LevelCtrl clusters", which combination should work?

Switch type - momentary
Relay mode - short press (or detached, if needed)
Binded mode - short press

clumsy-stefan:
Ok, great! thanks!

When I use "short press" for binded mode, and I longpress the button, when does it send a onOff also or only when I short press?
And the other was round, when I select longpress, does it send both, a onOff and a levelCtr during long press?

I think these will then also specify the name to use for it, when it's clear what it does ;)

thanks again for your time implementing this!

romasku:
Only in specified mode. If it is "short press", then only after short press. If it is "long press", then only after long press.
And if you set it to "rise", then it sends On/Off command the moment you touch the switch, as it was before.

clumsy-stefan:
ok, and what happens on longpress when on rise?

besides what's the difference then of that switch (eg rise) instead of setting duration to 0? what should happen when duration is 0 on long press?

sorry, still trying to get my head around the logic...

romasku:
> ok, and what happens on longpress when on rise?
Here’s your text with grammar fixed and kept close to the original wording:

Maybe this helps:

* `on_rise` – perform the action when a press is registered.
* `short_press` – perform the action if the button is released before the long-press duration.
* `long_press` – perform the action if the button is held for longer than the long-press duration.

If `relay_mode` is set to `on_rise` and `binded_mode` is set to `long_press`, touching the button will trigger the relay immediately. If the button is then held long enough, the device will issue an **On/Off** command.

Conversely, if `relay_mode` is `long_press` and `binded_mode` is `on_rise`, touching the button will send the **On/Off** command immediately, and holding it long enough will trigger the relay.

If both are set to `on_rise`, touching the device triggers both actions.
If both are set to `short_press`, both actions trigger when the button is released quickly enough.
If both are set to `long_press`, both actions trigger after holding the button long enough.

Level control works independently of `relay_mode` and `binded_mode`. The defaults are configured so that the typical use case works correctly: `binded_mode` is set to `short_press`, meaning that when the device is touched, no command is sent immediately. Upon release, the device detects either a short press or a long press, and sends either an **On/Off** command or a level control command accordingly. However, it is still possible to configure it in any other combination.

clumsy-stefan:
Ok, I think I sbegin to understand.. thanks.

Just one more question for now:

> Level control works independently of relay_mode and binded_mode. The defaults are configured so that the typical use case works correctly: binded_mode is set to short_press, meaning that when the device is touched, no command is sent immediately. Upon release, the device detects either a short press or a long press, and sends either an On/Off command or a level control command accordingly. However, it is still possible to configure it in any other combination.

The "upon release" worries me a bit, because levelCtrl start should be sent immediately on long-press detect and stopped on button release... But I guess that's what you implemented?

I can live with that as I'm used to binary logic, but for the ordinary ZB user I think these possibilities/configurations are way too complicated to understand (even I had problems understanding and you had problems explaining, so that I understand)... but again, IMHO...

romasku:
> The "upon release" worries me a bit, ...

Sorry, bad wording. On a long press, it sends the "start" command, and on release, it checks: if it was a long press, send the "stop" command. Otherwise, handle the short_press targets, if any.

I found a bug: I was using the wrong constant to check whether the current press is a long press, which completely broke the short_press logic.
I’ve tested on a real device, and I’m now able to both turn on/off with a short press and use a long press for level control.

2000 ms for a long press is way too long -- I don’t know what I was thinking. What would be a more reasonable default?

romasku:
>  I think these possibilities/configurations are way too complicated to understand 

I agree that it’s complex and may require reading some documentation, but my hope is that it will “just work” with the default settings for most users (once I’ve fixed all the bugs, sorry about that). Also, with working device, it should be possible to understand what’s happening simply by trying it out.

clumsy-stefan:
>2000 ms for a long press is way too long -- I don’t know what I was thinking. What would be a more reasonable default?

I currently use 1000ms .... probably you could go back to 800 or a bit below... depends... for a levelCtrl 500-800 would probably be OK, for action on long-press you probably would have it a bit higher...

I'd also change the level-change-rate to 40 or even 50... most devices have  255 levels which means 5sec. when you set it to 50, already quite long...

And when talking about defaults, the "number of presses to reset" I would increase to at least 10, as with 4-gang switches, if you quickly turn on all 4 switches you only have one press left... it happens quite often as not everyone can remember which switch is for what, especially on 4-gang (even though it's written on them) so turning on and off again happens frequently... Until you find the correct switch you're quickly over the 5 presses...

clumsy-stefan:
Could do some tests tonight and basic functionality seems ok.

The only important change I'd see is to make "short press" default for the relay, otherwise when you long-press a button for levelCtrl and the button is on default "rise" then at the end of the press, when you release the button, the relay toggles (for momentary buttons), seems odd to me..

Generally as already mentioned, like it's now it works differently than any other system I know and came across and probably others that come from different technology won't understand...

But for my part I can live with this and I'm happy if you merge it.

Thanks again for implementing!

romasku:
> Generally as already mentioned, like it's now it works differently than any other system I know and came across and probably others that come from different technology won't understand...

Do you mean it is different on the configuration side or the user side? I thought it now works as others do on the user side... If it is not, let's improve it more. I'm okay if the configuration is "nerdy" as it is custom firmware after all, but it for sure should allow to configure a basic scenario. I thought basic scenario - press to toggle, hold to change brightness.

romasku:
> And when talking about defaults, the "number of presses to reset" I would increase to at least 10, as with 4-gang switches, if you quickly turn on all 4 switches you only have one press left...

Hmm, it is counted per button, not total for the device. So pressing all four of them shouldn't leave you with one press left. But I totally agree with setting this default higher, allowing it to be disabled if there is a physical reset button.

romasku:
>  the button is on default "rise" then at the end of the press, when you release the button, the relay toggles (for momentary buttons)

When `relay_mode = "rise"`, the relay should be triggered only at the start of the press. I just tested this on my device, and it worked as expected. Could you please double-check that with
`relay_mode = "rise"` and `binded_mode = "short_press"`, the relay is **not** triggered **after** releasing a long press?

Also, I’m curious about your setup — do you have a smart bulb that you control for brightness, but turn on/off via the relay (cutting power to it), or does it work differently in your case?

clumsy-stefan:
>Do you mean it is different on the configuration side or the user side? I thought it now works as others do on the user side...

Switching local (or remote) relays on longpress is just something others don't do... either it's a short-press and action or it's a long press action.. but not a long press combined with short press...

>it worked as expected. Could you please double-check that with
relay_mode = "rise" and binded_mode = "short_press", the relay is not triggered after releasing a long press?

Not sure what I did wrong. while double checking it's correct it switches on button press... But I think that's what I meant above, on long-press relays shouldn't trigger by default. But I'm absolutely ok with how it is... It's just different...

>Also, I’m curious about your setup — do you have a smart bulb that you control for brightness, but turn on/off via the relay (cutting power to it), or does it work differently in your case?

Currently only with smart-bulbs that are always powered. But I can imagine for powersaving that one could use the relay for powering the lamp ad longpress for levelCtrl.

>Hmm, it is counted per button, not total for the device. So pressing all four of them shouldn't leave you with one press left. But I totally agree with setting this default higher, allowing it to be disabled if there is a physical reset button.

Right... Probably I was just too nervous testing on-off and levelCtrl and triggered resets... but as you say, even for one button I think we should increase to 10...

>I thought basic scenario - press to toggle, hold to change brightness.

Absolutely.. but currently default on long-press is toggle relay AND levelCtrl... That's why I said I'd rather have it on short-press only...

romasku:
>  currently default on long-press is toggle relay AND levelCtrl.

Current default is relay_mode = short_press, binded_mode = short_press. Before the last change, it was relay_mode = rise, binded_mode = short_press, as I expected that if levelCtrl is used, then the user will set it to relay_mode=detached. I didn't intend any toggle to happen on long press, maybe there is some bug I'm missing?

clumsy-stefan:
So, my bad, sorry.. Probably I hadn't included the last change or not reset to the default.. But when it's short_press for both it should be reasonable, at least for the default behaviour!

I suggest to merge the PR and see what other users report, as it seems to run with basic functionality and not bricking anything!

clumsy-stefan:
PS: an additional comment... instead of the binding and relay modes (or in addition) I'd rather have the short- and long-press fucntion. Eg: long-press: levelCtl/colorCtrl/onOf/detached/etc... same for short press. like this we could easily add colorCtrl and the user could assign the different buttons to different functions for the same end device...

Or as a workaround you jus send both (levelCtrl and ColorCtrl) and the user needs to define himself what cluster he binds...

marazmarci:
Hey, everyone! I've been following the threads about the `genLevelCtrl` topic for a while, because I think it's the biggest enhancement that this FW needed, while I agree with @clumsy-stefan that this FW is "the missing piece" for a perfect Zigbee light setup. 👌 Thanks for the good work on it!

I have a completely different idea. Sorry for only posting it now, when two of you have already opened a PR, but I _just_ got this idea. 🤷 😄

Add separate endpoints for the different button press types. For example:
- `switch_left_short_press`
  - `OnOff`
- `switch_left_rise` ✳️
  - `OnOff`
- `switch_left_long_press`
  - `OnOff`
  - `LevelCtrl`
- `switch_middle_...`
  - ...
- ...

This would eliminate the need for complex configuration parameters, open up more binding use cases, and make the binding configurations more like a single source of truth. One new use case that this would enable is toggling another switch or light with a long press.

This would certainly affect the size of the binding table again, which was increased in #100.

**What do you think? Is this a viable approach, or am I missing something?**

✳️: I think `rise` is too technical of a term. Even though this firmware is for tech-savvy people, we can't expect everyone to know what a _rising edge_ is. For those who don't know, it can mean the exact opposite, such as when the button physically _rises_ up from its pressed state when released. 😄
I think It would be more user-friendly to use these terms:
- `press_start`
- `short_press`
- `long_press`

clumsy-stefan:
@marazmarci thanks for the input! Basically I like the idea (especially of adding functionality)..

I also already thought about adding individual endpoints (I think we already discussed that a bit earleir on), but I fear that it then gets really complicated in the binding table. The enduser would need to know which endpint corresponds to which action. For a 4-Gang switch you would end up in some 30 Endpoints (buttons, relays, etc.). Also we would need to think about multi-presses and how they are handled..

One usecase which will be difficult to implement is if you want to use the "left" switches for off/dimdown and the right ones for on/dimup as it's done in quite a lot of commercially available switches...

Similar if you want to include color-control, you need to think about what exactly you like to send.

IMHO you should/could look from the other side; for each button action (short-press, long-press, multi-press, probably release) you can define which action it should trigger. Eg. "start dim up/down", "start dim down", "on", "off" "toggle", "shoot your cat", "stop dimming", etc.. whatever you can think of and implement as a ZB function.

If you're bold enough you then could add actions as "double-press", "short-long-press" etc.

This would end up in a 2-dimensiona matrix of "available presses" with "available functions" for ech button. Probably you would need to have a second option per button for "local action" and "remote action" due to local builtin relays.

If that's feasible with the current UI's (eg. z2m) I'm not really sure...

However, having said that I need to revert to my own principle and praises to KISS.... While all these ideas are a real nice to have (and interesting from a coding pov), I think 95% of the users neither need it nor will (know how to) use it... Have you ever been somewhere, where one switch/button controls multiple endpints?
That's why I would vote to implement the basic "short-press -> onOFF", "long-press -> levelCtrl" with a single Endpoint now... When that works probably add "coloCtrl" for longpresses which will be automatically selected by the correct binding..

If that's widely in use we can start to think of how geeky we want to make it.. Otherwise we'll never reach a production state and try to implement all cool possibilities and fix numerous issues before it's in real use; and in that complexity there will be a lot of issues, which will need a lot of time (will @romasku have that time and willing to invest?)... (again IMHO)...

So not meant to offend anyone, but that's just my 2cents...

romasku:
Hello,

> Add separate endpoints for the different button press types. For example: ...

I like the idea from a technical perspective, but Z2M and especially ZHA lack proper UI support for this. In ZHA, binding something is already a challenge because it doesn’t provide named endpoints. Even in Z2M, named endpoints don’t always work for me, so I often fall back to using the endpoint number. Having many endpoints would only make configuration much more difficult, so I think the benefit of supporting such cool but somewhat extreme use cases isn’t really worth it.

> I think rise is too technical of a term.

I completely agree, and I’ll implement it. Thank you for pointing this out — I had gotten used to it and stopped noticing it was a problem.

P.S. I’m planning to finish and merge it tomorrow, so this is your last chance to suggest any changes:)

PR #116, opened 2025-08-10
Fixed config string length issue + add known_issues doc

andrei-lazarov:
Moved here from #89


romasku:
Thank you!

clumsy-stefan:
One question (before I actually looked depepr through it): shouldn't we take this opportunity (where most people prbably need to completely reset the device anyways)  to move the config from `NV_MODULE_ZCL` to `NV_MODULE_APP`?

romasku:
I think we can automate that migration using the mechanism you introduced in https://github.com/romasku/tuya-zigbee-switch/pull/109

andrei-lazarov:
Resetting shouldn't be necessary for that.. as I added lastSeenVersion. Correct me if I'm wrong.
I can try that out before pushing the update if you wish.

But if it does require resetting, I suggest we do the switch once we also have the unique binary (without the pinout hardcoded). I don't think we have a way to tell all users: reset before updating.

Right now resetting is not needed.. only recommended if you are using a device with a custom config.
That should be very few cases.. we quickly added all devices to the databases.

romasku:
Also, I'm releasing this as 1.0.18, so that this fix is published to all users.

clumsy-stefan:
BTW: creating a unified binary is quite easy already now by only amending the device-db and using two converters in `switch_custom.js`, one fot the "clean" build with only `device-config` in it and one for the 4-Gang Switches (currently the only ones I have) with everything in it...

Of course this only works, as I have the first version of the config-string fix included from @andrei-lazarov ... Will try next week with a clean build and the new fix..

EDIT/PS: my main issue lies in #106 as leaving and rejoining always reverts back to the "clean" FW.... 

I currently use the following 3 entries for all my devices:
```
TS0004_CF:
  name: TS0004 generic 4-Gang
  config_str: Custom-FW;TS0004-CF;
  device_type: router
  tuya_manufacturer_id: 4417
  tuya_image_type: 54179
  firmware_image_type: 43999
  stock_converter_model: TS0004_switch_module
  tuya_manufacturer_names:
      - _TZ3000-CF // only needed to make sure the create readme helper script doesn't fail
  old_zb_models:
      - 5ajpkyq6
      - mmkbptmx
      - ltt60asa
      - 5ajpkyq6vv
      - knoj8lpk
  human_name: Generic TS0004 4-Gang Switch
  status: Experimental
  github_issue:
TS0004_CFM:
  name: TS0004 generic 4-Gang
  config_str: Custom-FW;TS0004-CF;
  device_type: router
  tuya_manufacturer_id: 4417
  tuya_image_type: 43999
  firmware_image_type: 43999
  stock_converter_model: TS0004_switch_module
  tuya_manufacturer_names:
      - _TZ3000-CFM  // only needed to make sure the create readme helper script doesn't fail
      - 5ajpkyq6
      - mmkbptmx
      - ltt60asa
      - knoj8lpk
  human_name: Generic TS0004 4-Gang Switch
  status: Experimental
  github_issue:
CLEAN_CF:
  name: Generic clean FW
  config_str: Custom-FW;CLEAN-CF;
  device_type: router
  tuya_manufacturer_id: 4417
  tuya_image_type: 54179
  firmware_image_type: 43999
  stock_converter_model: Various
  tuya_manufacturer_names:
      - _TZ3000_5ajpkyq6
      - _TZ3000_mmkbptmx
      - _TZ3000_ltt60asa
      - _TZ3000_5ajpkyq6vv
      - _TZ3000_knoj8lpk
  old_zb_models:
      - _Tuya-TS0004-Avatto-custom
      - TS0004-Avatto-custom
      - TS0004-Avv
      - _TZ3000_4channel
  human_name: Generic TS0004 4-Gang Switch
  status: Experimental
  github_issue:
```

After that I use the following config strings for the devices (some examples):
```
5ajpkyq6;TS0004-CF;BC2u;LD2i;SD3u;RC0;SD7u;RD4;SB6u;RC1;SA0u;RC4;
mmkbptmx;TS0004-CF;BB6u;LB1;SC1u;RB7;SC2u;RB5;SC3u;RB4;SD2u;RC4;
knoj8lpk;TS0004-CF;BC3u;LC2i;SB5u;RD2;SB4u;RD3;SD7u;RC0;SD4u;RC1;
```

clumsy-stefan:
When changing the config-string the device goes dead. only after a powercycle/reboot it talks again to the cordinator... At least in my environment...

It seems to die when doing the interview...

andrei-lazarov:
Thanks again @clumsy-stefan for stress-testing every new change.

On the current main branch, I am able to reliably switch between 1, 2, 3 gang config strings and have the device work accordingly.
But, as always, the 4 gang config string is still problematic..  
Thankfully, this time it only freezes the device and it recovers after a reboot.

I'm sorry I did not catch the issue before merging..
But this is still a lot better than immediate bricking.. so we should keep the patch.

Let's continue discussion on #77 as this thread is marked as closed. I'll detail the behavior there.

Issue #110, opened 2025-08-07
Change/standardise ZB-Model and ZB-Manufacturer definitions


clumsy-stefan:
As discussion started in #105 and requested by @romasku I open a separate issue for this.

The goal is to change some information in the devices to simplify and standardize the device specifications while keeping a reference to the original firmware/device.

My proposal would be that:
- the `Zigbee Model` includes the unique Name of the converter that need to be used. Basically one for each 1-2-3-4 Gang switch, probably a handfull more. 
- the `Zigbee Manufacturer` contains a reference to the original firmware/swich. Eg. the second part of the original Manufacturer string to keep it short.
- `Image-Type` stays as it is (unique FW-Identifier)

Advantages:
- the number of converters can be kept at a minimum. Same type of switches use the same converter
- standardize the naming
- a reference to the original firmware is kept and can be uniquely identified
- firmware updates can be uniquely identified from the Manufacturer and image type fields combination

As an example, a Tuya 4-Gang Switch with a orignial Manufacturer `_TZ3000_knoj8lpk` that uses a `TS0004_switch_module` would become:
- `Zigbee Model: TS0004-CF` (used for standard 4-Gang converter)
- `Zigbee Manufacturer: knoj8lpk` (original second part of MF, used for identifying the exact type of the switch)
- `Image-Type: 43548` (uniqe custom FW type used for identifying the correct binary with the correct device configuration string)

I did not yet look further into how this could be implemented, but I think a more or less simple change of the `device-db.yaml` could do it (@andrei-lazarov: probably you could help ;) ). Just need to make sure we're backwards compatible.

andrei-lazarov:
I like the idea.. we should optimize manufacturer and model names.
It can totally be done with backwards compatibility (Thanks romasku for thinking of everything 😃).  
I can do it myself.

And even if we keep the full original manufacturer name, we still have the fw image type to differentiate between stock fw and custom fw.

I can see why romasku made custom names for every device. It gives you the possibility to customize each device on its own.
But the devices are so similar.. there's no need. In the future we can revert to unique names if we need to.

clumsy-stefan:
> I can do it myself.

Thanks for your help!!!

> And even if we keep the full original manufacturer name, we still have the fw image type to differentiate between stock fw and custom fw.

The problem is the string legth I guess/thought? That's why I proposed to make it shorter.. (and it seems the TZ3000 is the same anyways in the newer models)

[...]

PR #118, opened 2025-08-14
Clean database and converters

andrei-lazarov:
I am working on cleaning the database as discussed in #110:
- [x] show correct Z2M model and picture with the rebranded model names (overwrite with whiteLabel)
- [x] remove duplicate name (the name field is not used anywhere, we already have the name in the key and in the human_name field)
- [ ] automate some processes so we minimize the information needed in the database (eg. get the whiteLabels from original Z2M converters)
- [x] use original manufacturer name in the custom firmware so we keep a reference to the original device
- [ ] ~~standardize the converters and reuse them~~
- [x] add Zigbee board in the database (ZTU, ZT3L etc) - seems useful, I would use it in the converters to prevent impossible pinouts
- [x] add AliExpress link in the db?
- [x] split `tuya_manufacturer_names` into tuya name and old fw names
- [x] combine router and end_device entries

I will do each task in a separate commit so it's easier to review

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -37,11 +37,18 @@ jobs:
 
     - name: Build for all boards
       run: |
-        yq -r '(keys)[]' device_db.yaml | while read ITER; do
-          echo "Building for board: $ITER"
-          BOARD=$ITER make clean && BOARD=$ITER make -j16 || exit 1
-          echo "Checking if files were created for board: $ITER"
+        yq -r 'to_entries | sort_by(.key)[] | "\(.key) \(.value.device_type)"' device_db.yaml | while read ITER TYPE; do
+          echo "Building for board: $ITER (router)"
+          BOARD=$ITER DEVICE_TYPE=router make clean && BOARD=$ITER DEVICE_TYPE=router make -j16 || exit 1
+          echo "Checking if files were created for board: $ITER (router)"
           ls -l bin/$ITER/
+
+          echo "Building for board: $ITER (end_device)"
+          if [ "$TYPE" = "end_device" ]; then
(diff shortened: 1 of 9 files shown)


romasku:
>  standardize the converters and reuse them

I'm currently working on a solution to generate z2m "exposes" based on device config directly inside the converter. Idea here that changing the config should update the Z2M UI, and should not rely on the exact model/manufacturer string. So I'm not sure you should do this change now, as my change will probably be in conflict.

andrei-lazarov:
That sounds interesting.. didn't know we could do that! Does that mean we'll have a single converter that adapts to every device?

romasku:
> Does that mean we'll have a single converter that adapts to every device?

Yes! I finally got it working. It still needs some polishing, but hopefully I’ll open a PR tomorrow.

romasku:
Here is the generic converter: https://github.com/romasku/tuya-zigbee-switch/pull/120

romasku:
> automate some processes so we minimize the information needed in the database (eg. get the whiteLabels from original Z2M converters)

Maybe you have some additional ideas in mind here, but regarding whiteLabels from Z2M, I don’t think that’s an option for us -- in addition to different branding, white labels often have different pinouts.

andrei-lazarov:
I just wanted to automatically get the value for override_z2m_device.  

Currently we have to manually search through official Z2M converters and add it in the database (to get the correct device name and picture).
So nothing pinout related. Since you already reviewed, I'll do it later in another PR.

**Glad you like the changes! Please merge this.**
Also go through your devices (when you have the time) to fill in the zb_module.

romasku:
Thank you!

Issue #128, opened 2025-08-28
Support for _TZ3000_tqwydnqn


rastakozza:
Hi,

I got a 3 gang wall switch zigbee model _TZ3000_tqwydnqn.

<img width="484" height="306" alt="Image" src="https://github.com/user-attachments/assets/b02e64cb-756e-442e-9a6a-ee39c888a476" />

<img width="534" height="287" alt="Image" src="https://github.com/user-attachments/assets/873036dd-7749-4a3d-adc4-137300165071" />

<img width="693" height="632" alt="Image" src="https://github.com/user-attachments/assets/e4fe962a-da79-4080-870f-c9db7681317d" />

I cloned _TZ3000_qewo8dlz entry in index_router-FORCE.json:

```
  {
    "fileName": "1141-aa0d-ffffffff-tlc_switch-forced.zigbee",
    "fileVersion": 4294967295,
    "fileSize": 166418,
    "url": "https://github.com/romasku/tuya-zigbee-switch/raw/3fa33d1fb514f7f6a4ccf609c425198992882f0e/bin/MOES_3_GANG_SWITCH/1141-aa0d-ffffffff-tlc_switch-forced.zigbee",
    "imageType": 43533,
    "manufacturerCode": 4417,
    "sha512": "eb9199afe7cdc3bd6fde8f803c4ae7978680bba1875dd492ab5a8dde842851f3a404cc03f25abd079212d308e7addd528b249ce513674f1fe4028faa8a42af7e",
    "otaHeaderString": "Telink Zigbee OTA\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000",
    "manufacturerName": [
      "_TZ3000_tqwydnqn",
      "tqwydnqn"
    ]
  }
```

I see the device in OTA tab, but no update are available.
As specified in the porting guide, here is the log line:

```
[...]

PR #129, opened 2025-08-28
Add _TZ3000_tqwydnqn

andrei-lazarov:
Added **Manhot 3-gang button switch with optional Neutral** from #128


andrei-lazarov:
Device added to main in #155, but not building it yet.

Issue #130, opened 2025-08-29
Support for _TZ3000_imaccztn


bigroma73:
Hi
I've got 4 channels Zigbee relay 
_TZ3000_imaccztn

Does it possible to support this relay? 

WBR
Oleksandr

![image](https://github.com/user-attachments/assets/35e6c737-a5b9-4675-a566-9728651b69f9)

![image](https://github.com/user-attachments/assets/61b9ec1e-126a-40de-bfa4-8d4eaffd7a8a)

andrei-lazarov:
I have not seen this form factor before, but it should be possible.  
I assume it works like a normal switch. And the ZT3L is supported.  

So you have to trace the pinout with a multimeter.  
Look up the [ZT3L diagram](https://developer.tuya.com/en/docs/iot/zt3l-module-datasheet?id=Ka438n1j8nuvu) and let me know what's the pin for: every button, the reset button, every relay, state indicator LEDs and network LED. How many LEDs does it have?  
Note that the diagram shows the back of the module. And you are viewing it from the front.

Also, I'm curious what you are using this for and what features from the custom firmware you would need.

bigroma73:
I've found buttons pins on module but now I still not found relays pins. They're connected not direct. 

I'm using switches like this for different applications: gate, watering, water tanks. This one and one like this from Avatto(same issues) I'm using for 24V AC watering in my garden. On enabling channel sometimes other channels disabled.

bigroma73:
There's additional chip between ZT3L and transistors. Buttons K# connect directly K1:D7, K2:C0, K3:B5, K4:RX, but mode button, blue led and relays connect to 20 pin chip marked SMG(or SMC)-C45.
I can't googled this chip :( 
![image](https://github.com/user-attachments/assets/8472bbe3-7163-4cfa-9653-77dd7db6ad3f)

andrei-lazarov:
Well, you can also test the voltage of the pins with the board powered on. I see it has a microUSB input, so 5V should be safe to handle.  
I would power it on, then toggle the relays from ZHA/Z2M and check which pin changes its value.

bigroma73:
Hi
Looks like
C3:RedLED
D2:relay3
C2:relay4
D4:relay1
A0:relay2

I can't find blue led and reset/mode button :(

andrei-lazarov:
I assume the red LED is the network indicator and there are also 4 blue LEDs indicating the relay states?  

[...]

PR #133, opened 2025-08-30
Add TS0004_MHCOZY

andrei-lazarov:
Added MHCOZY TYWB 4ch-RF - 4 channel board with microUSB from #130.
Note that the factory firmware has **On with timed off**, a feature we are currently missing.


andrei-lazarov:
Added in #155

PR #132, opened 2025-08-29
Touch-up readme, docs, changelog

andrei-lazarov:
Improve readability and add some information here and there

PR #137, opened 2025-09-02
Add support for off_with_effect

andrei-lazarov:
This is on top of #132 

Mishakov found out that some remotes use the off_with_effect command instead of the usual off command.

So I added it to `relay_cluster.c`, along with on_with_recall_global_scene.


romasku:
The change looks good, thank you. However, let’s merge the readme/docs touch-up separately first.

andrei-lazarov:
Fixed. Sorry for the mess of PRs 😃

romasku:
Super, thank you very much!

Issue #127, opened 2025-08-27
Control onboard LED


soosp:
I have several Avatto LZWSM16-W2 modules. In the dark, it is a little distracting that the light from the device's onboard LED shines through from behind the switch. Is it possible to control the onboard LED?

andrei-lazarov:
It is possible. On Z2M there is a toggle to control the LED.  
It looks like it's missing from the ZHA quirks though. We need to implement it.

andrei-lazarov:
@soosp Fixed. Update your quirks and close the issue after you test it 🙂 
Sorry it took so long... for some reason I lost motivation half way through

soosp:
It works as expected 🙂 Thank you! It seems that your motivation came back now 🙂

Issue #63, opened 2025-06-28
Avatto ZWSM16-2 ZHA quirk support


soosp:
I just updated an Avatto ZWSM16-2 switch and paired with ZHA. It introduce itself as TS0002-AV-CUS, but there is no such entry in `switch_quirk.py`.
The basic switching functionality is working, but there are neither enhanced config options nor power metering. Fortunately I do not need them, just report it.

soosp:
Finally, adding line 
```
"TS0002-AV-CUS;TS0002-AV-CUS;BC2u;LD2;SD3u;RC0;SD7u;RD4;",
```
made it work.

andrei-lazarov:
Thanks for pointing this out. The ZHA quirks were not updated.  
I made a pull request that will keep them up to date: #69

soosp:
The entry above is missing from the new quirk.

andrei-lazarov:
We changed the device model name from AV-CUS to AVB.  
If you update your device to the latest version, interview and reconfigure it, it will receive the new name.  
Though Z2M converters keep old names for backward compatibility.. We should make ZHA quirks do the same.  
Thanks for letting us know!

soosp:
Thank you! The device was removed and re-added and recognised on the new name now.
BTW: Are there any concept on device namig?

andrei-lazarov:
Ah.. ZHA does not have the dedicated interview and reconfigure buttons, right? Indeed that means you have to remove and re-add it to the network to use new features.  

And removing it erases user configuration (switch type, action mode etc) so that must be very annoying.. We have investigated this and did not succeed with keeping user config after leaving the network.

I must check ZHA and update the documentation.

Regarding names, we just need to keep them as short possible. We are having issues with long strings (needed for device config string)

soosp:
In ZHA there is Reconfigure option, but it do not change the model name and manufacturer. I assume it use IEEE for identify the device.

andrei-lazarov:
The interview button updates manufacturer name and model name. The reconfigure button updates endpoints and reporting. In that case you're good, we don't change names often. And adding the old names to the quirk should be easy

andrei-lazarov:
@soosp The ZHA quirks now contain all combinations with old and new names.
Update your quirks and you may close this issue. Thanks again for reporting!

soosp:
It works well for me. Thanks for your efforts!

PR #141, opened 2025-09-03
Improve ZHA quirks

andrei-lazarov:
There's a bunch of things missing from the ZHA quirks (that we already have in Z2M converters):

- [x] backwards compatibility #63 
- [x] network LED #127 
- [x] toggle_smart action modes
- [ ] device config string 
- [ ] corresponding switch/relay number in the names
- [ ] descriptions (if possible)
- [ ] possibly more stuff

So I'll try to add them for our ZHA friends 😃

I am running a temporary instance for testing. So far backwards compatibility and network LED work.  

The LED control currently shows up under Configuration (with all device options).  
I think it would be better under Control (with the switches), but I can't figure out how to move it there.


romasku:
You can use this option in `QuirkBuilder.switch` to select proper category:

https://github.com/zigpy/zigpy/blob/feef7fa4b56ca35fe841e48e4c3592fcbdbaeb2c/zigpy/quirks/v2/__init__.py#L963

So like 
```
.switch(
            CustomBasicCluster.AttributeDefs.networkIndicator.name,
            CustomBasicCluster.cluster_id,
            translation_key="network_indicator",
            fallback_name="Network indicator",
            endpoint_id=1,
            entity_type: EntityType = EntityType.STANDARD,
)
```

Didn't test it, but it should work.

andrei-lazarov:
**Adding toggle_smart and renaming the custom entities was easy.**

Then I cloned the zigpy repo to see how QuirkBuilder works.
But I couldn't figure out how to rename the 'normal' entities: relay state and power-on behavior.  
These also work without the quirk, but show up as Switch, Switch, Switch. I'm curious if it looks the same on original firmware..
Anyway, it does not matter. **The users can click and rename the entities to whatever they want.**

**Adding the config string is not possible as far as I can tell**. No option for string in the builder.
Descriptions also not possible in ZHA.

Also, changes are non-breaking. The entity labels stay the same, even if I changed the names. 
They only update if you manually click the update button. Also you wouldn't change config options in automations..

**Let's merge it like this.**

andrei-lazarov:
Reverted to draft because the LED is actually undesirable in the 'Controls' section. 
Having it there means it gets toggled when you toggle the whole device. (Toggling the whole device is expected to toggle only the relays)
The Z2M converters need to fix this too. There every config option shows under 'Controls'.

andrei-lazarov:
Completed now

romasku:
Great job as always, thank you!

PR #148, opened 2025-09-14
Update documentation 😭

andrei-lazarov:
I thought the updating guide was comprehensive, but it has caused confusion for multiple persons..  

I got some feedback and made the following changes:
- info about the Z2M converters path!! It depends on installation method.
- moved the install instructions at the top of the page so people see them right away!
- explained the last update step better
- added more section links
- added index link format
- added recommended indexes
- noted that things are branch specific
- added info about faster updates
- emphasized Zigbee manufacturer
- added warning about ZHA
- updated device count
- added sockets to the description as we'll support one soon
- updated known issues
- categorized docs into sub-folders (and edited all the links... and checked if they work...)

I also added a section about the official converters, as we are working closely with them.  
It feels like we rush to install the custom firmware, before fully supporting the stock firmware.  
I hope you don't mind it being a little off-topic and the the page didn't become too long for the users...


romasku:
Thank you very much! Great job!

PR #154, opened 2025-09-18
Add build flag in db

andrei-lazarov:
I added the **`build`** field in the database:
- **yes**: builds fw and everything for that device
- **no**: skips building for that device
- **null** (missing): default to yes

**This allows adding incomplete / in-progress devices in the database** (and potentially in the devices table).  

For example, some stuff I want to add:
- I have the pinout for the 3-gang iHseno touch switch, but I don't know the manuf name yet
- Compatible devices that were not traced yet
- Devices that are unstable because we didn't fix bi-stable relays yet

Changes:
- edited all the scripts and workflow to skip building incomplete devices
- unrelated: fixed index script crashing if you renamed the git remote ('origin' name was hardcoded)
- updated db explanation: `build` flag + missing `stock_converter_manufacturer`

I tested both locally and online, with all values for the build flag: yes, no, missing.  
Everything looks good, should be safe to merge.


andrei-lazarov:
Rebased into main so no more conflicts. 
I tested local and online builds again, look good.

romasku:
Thank you!

PR #158, opened 2025-09-27
Update documentation

andrei-lazarov:
Mainly added more recommendations, tips, fixes here and there.  
Also updated the db explaining doc to match #155.  
Safe to merge.


romasku:
Thank you!

PR #160, opened 2025-10-08
Normally closed momentary buttons

andrei-lazarov:
Some guy on Reddit needed support for normally-closed switches, so I started working on it. 
I'm surprised we went so far without them. Also I haven't seen any Zigbee device that has this feature on original firmware. 

It's a little hard for me to test, as my switch wires snapped off.

But so far press and release work. I'll probably figure out long_press in a bit.


andrei-lazarov:
Reverted to draft, as the reset by spamming the switches is triggered too quickly/often? Maybe the counter is never reset. I'll check it out

andrei-lazarov:
For momentary NO switches we need to reset the multi-press counter on press.
For momentary NC switches we need to reset it on release. (their press)
For toggle switches we need to reset it on both.

Tested and works well now.

Also for NC we should count 'releases' (their presses). But currently it does not matter for the spam reset method, so I'm making as little changes as possible to prevent new bugs. 
However `button.c` does need some refactoring for the function names to make sense for all kinds of buttons. And some changes to allow double-press, triple-press.. I will revisit this later.

andrei-lazarov:
I also added **action states for toggle buttons:** `position_on`, `position_off`. I used this PR to avoid conflicts. 

It used to show `pressed` and `released` (the states from momentary switches). 
So naming a little wrong, but they were usable. However, I think they got removed with the levelCtrl update and people have since been asking for them.. #168, @Yokosa on Discord.

This should bump the version to 21

romasku:
**Great work, thank you!**
Your code is clean and very readable -- it was a pleasure to review :).

To make the button implementation less complex, we could allow configuration of which logical level should be considered “pressed” or “released.”
This way, most of the code would remain unchanged, with only a small modification like this:

```c
void btn_update(button_t *button)
{
    u8 is_pressed = drv_gpio_read(button->pin) == button->pressed_gpio_level;
    ...
}
```

Of course, this requires adding a new field to the `button_t` structure:

```c
typedef struct
{
    u32 pin;
    u8  pressed_gpio_level;
    ...
} button_t;
```

Then both `config_parser.c` and `switch_cluster.c` should set this field appropriately.
And, to update the value on Zigbee writes, we could do something like:

```c
void switch_cluster_on_write_attr(zigbee_switch_cluster *cluster)
{
    if (cluster->mode == ZCL_ONOFF_CONFIGURATION_SWITCH_TYPE_MOMENTARY_NC) {
        cluster->button->pressed_gpio_level = 1;
    } else {
        cluster->button->pressed_gpio_level = 0;
    }
    switch_cluster_store_attrs_to_nv(cluster);
}
```

**That said**, I don’t think the current implementation is too bad to merge as it is -- it’s already clean and readable.
If you prefer, I can merge it now and leave this refactoring for a future improvement.

andrei-lazarov:
That's a great suggestion that makes my code look foolish. Thank you!  
I will definitely use that to refactor the code, but I think it's best to merge this for now.. so people can use the new features right away.
I am looking forward to refactoring and adding double clicks while I'm there.

falkenhawk:
@andrei-lazarov thank you very much for this! What modules do you know that send low voltage on COM so that S1, S2 ... button input may be fed with that low voltage, instead of live 230V? I'd like to use this feature for a wired magnetic sensor in a door - when the door is open, the connected light should turn on.

andrei-lazarov:
@falkenhawk that's a very good idea! 👍🏻 

I'm also not a fan of battery-powered devices, so I wanted to DIY a motion sensor. 
But you're right, I can just use this firmware and connect the sensor to the switch wires - saving me the hassle of finding a device and writing firmware again.  

Unfortunately, none of the Tuya devices have galvanic isolation. 
So AC Neutral is the reference point for DC ground.

My understanding of this (over-simplified):
- if you plug it right    (L→L, N→N), the DC circuit operates at 0-5V (but could be higher).
- if you plug it wrong (L→N, N→L), the DC circuit operates at 230-235V.
And indeed I noticed arcs at the button inputs when I tested the wrong wiring.

Button inputs work by bridging COM (which is both DC ground and AC Neutral) to the corresponding pins: S1, S2 etc. 
Theoretically this is safe if wired correctly.. I don't have better options.

andrei-lazarov:
@falkenhawk Actually you can find some USB-powered devices:
- https://github.com/romasku/tuya-zigbee-switch/issues/130 
https://www.aliexpress.com/item/1005005457296054.html
- https://www.aliexpress.com/item/1005007598874816.html
- https://lilygo.cc/products/t-zigbee-esp32-c3-tlsr8258 ??

PR #162, opened 2025-10-08
🔒️ Make downloading the toolchain more secure

marazmarci:
I noticed that a HTTP link was used for downloading the toolchain. I changed it to HTTPS:
`https://shyboy.oss-cn-shenzhen.aliyuncs.com/readonly/tc32_gcc_v2.0.tar.bz2`

For me, the link seemed a bit fishy at first glance, but according to VirusTotal, the file is not harmful:
https://www.virustotal.com/gui/file/33b854be3e3db3dba4b4dacdda2cd4ea1c94dfd4d562864a095956de7991b430

Just to make sure, I also added checksum verification when the archive is downloaded. If the verification fails, the file will be deleted and the `make` target will fail.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -278,7 +278,7 @@ sdk:
 toolchain:
 	mkdir -p $(TOOLCHAIN_DIR)
 	wget -P $(TOOLCHAIN_DIR) https://shyboy.oss-cn-shenzhen.aliyuncs.com/readonly/tc32_gcc_v2.0.tar.bz2
-	echo "33b854be3e3db3dba4b4dacdda2cd4ea1c94dfd4d562864a095956de7991b430  $(TOOLCHAIN_DIR)/tc32_gcc_v2.0.tar.bz2" | sha256sum -c -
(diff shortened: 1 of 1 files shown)


andrei-lazarov:
This is cool! Thanks for the fix and check. 
I would have suggested we upload it to a safe place, but redistributing is probably not allowed.
Also I'm assuming it's extracted from [Telink IoT Studio](https://wiki.telink-semi.cn/wiki/IDE-and-Tools/Telink_IoT_Studio/) so we can use whatever file editor? Maybe you can also check that, to see if it's the same file.

marazmarci:
> Also I'm assuming it's extracted from [Telink IoT Studio](https://wiki.telink-semi.cn/wiki/IDE-and-Tools/Telink_IoT_Studio/) so we can use whatever file editor? Maybe you can also check that, to see if it's the same file.

You were right, I checked and it seems to be extracted from it, as all the files in the `tc32` folder match, except for a few ones which are only present in Telink IoT Studio (namely two folders: `addontools`, `include`). :heavy_check_mark:

marazmarci:
(in my latest push, I just removed an accidentally created merge commit)

romasku:
Thank you for hardening this part!

PR #163, opened 2025-10-08
✏️ Fix typos and improve grammar

marazmarci:
First of all, thank you @romasku and @andrei-lazarov for creating such cool documentation! :clap: :memo:

After reading through all of it, I found a few typos and potential improvements.

A more awkward thing I've found is that the word "binded" is used throughout the project, but the correct form would be "bound", and it is used even in the converter, so there's no going back (without a breaking change), so I didn't change it. I guess it's now part of the dictionary :smile:


romasku:
Thank you very much for fixing typos and mistakes!

Regarding "binded" -- I'm aware of this. If we have to make some breaking changes in the converter, we'll fix this as well. Thanks again.

PR #165, opened 2025-10-09
🔨🐛 Fix None model in Z2M converters

andrei-lazarov:
Looks like I missed something in #155.
Filling override_z2m_device with `null` resulted in `None` models in the converters.
(So Model: None and no picture in Z2M)

Safe to merge.


romasku:
Thank you!

PR #169, opened 2025-10-09
Fix Moes switch 1gang name and icon in Zigbee2mqtt

kernelzru:
There was change in recent commin to Zigbee2mqtt converters. Fix name & icon for MOES switch 1gang


andrei-lazarov:
Yeah, sorry.. my bad. I did come up with the fix here: #165 so you can close this

kernelzru:
fix in https://github.com/romasku/tuya-zigbee-switch/pull/165

PR #170, opened 2025-10-11
Add identify cluster, On with timed Off command, and Scenes

ThatMishakov:
This should add full support for scenes (saving OnOff attribute of relay clusters), identify cluster (needed for touchlink and I think it's mandatory) on relay endpoints, scenes, Basic cluster on all endpoints, and set the right simple description device IDs for the relay clusters.

The identify cluster either toggles the relay output, or blinks the indicator LEDs on the switches that have them. Since I only have relays, I haven't been able to test that, so if someone could test it, it would be appreciated.

Some considerations:
- I haven't tested it on 4 gang switches, but I think the default 32 clusters setting might not be enough on devices with many inputs and outputs (ZCL_CLUSTER_NUM_MAX in stack_cfg.h). This changes it to 48.
- ~~I think that 8 scenes is very little for switches, since if I'm not mistaken, each endpoint uses 1 table entry per scene, so in a 4 gang switch you will be left with 2 scenes if you want to control all the 4 outputs (ZCL_SCENE_TABLE_NUM). I have tried it with the size of 32, and it seemed to work just fine, so perhaps that could also be adjusted (?)~~
- The spec is not very clear to me, but I think identify is also mandatory on the switch endpoints, but I don't really know what would be the best way to implement it.
- ~~On a separate note, I think Basic cluster is also needed for each endpoint (I have it implemented in a different branch together with touchlink changes, but I couldn't get touchlink to work yet...)~~ I've changed my mind, I think this can also be added (?)
- All of my relays are running on the new SDK, so I haven't really tried the main branch with these changes, besides flashing one of my test relays

### DIFF
--- a/src/configs/stack_cfg.h
+++ b/src/configs/stack_cfg.h
@@ -49,7 +49,7 @@
  *  @brief  ZCL: MAX number of cluster list, in cluster number add  + out cluster number
  *
  */
-#define ZCL_CLUSTER_NUM_MAX        32
+#define ZCL_CLUSTER_NUM_MAX        48
 
 /**
  *  @brief  ZCL: maximum number for zcl reporting table
@@ -61,7 +61,7 @@
  *  @brief  ZCL: maximum number for zcl scene table
  *
  */
-#define ZCL_SCENE_TABLE_NUM        8
+#define ZCL_SCENE_TABLE_NUM        32
 
 /**
  *  @brief  APS: MAX number of groups size in the group table


src/device_config/config_parser.c
```diff
@@ -222,25 +227,33 @@ void parse_config()
...
   {
+    zigbee_endpoint_init(&endpoints[index], HA_DEV_ONOFF_SWITCH);
     switch_cluster_add_to_endpoint(&switch_clusters[index], &endpoints[index]);
   }
   for (int index = 0; index < relay_clusters_cnt; index++)
   {
+    relay_clusters[index].scene_cluster = scene_clusters + index;
+    scene_clusters[index].relay_cluster = relay_clusters + index;
+
+    zigbee_endpoint_init(&endpoints[index], HA_DEV_ONOFF_OUTPUT);
```
marazmarci: When I built your branch, it bricked my 3-gang device, and I went through the commits and found that commit 07a29708 caused it. I also found a fix for it, and now my device boots up. (Disclaimer: I didn't yet test it further than that)

I found two problems:

1. The endpoint `index` passed to the second call to `zigbee_endpoint_init` needs to be incremented by `switch_clusters_cnt`, so the same endpoint doesn't get initialized twice.
2. Clusters cannot be added to uninitialized endpoints, so the `zigbee_endpoint_init` calls need to be moved before adding the clusters.

Here's a suggestion that fixes these issues:

```suggestion
  for (int index = 0; index < switch_clusters_cnt; index++)
  {
    zigbee_endpoint_init(&endpoints[index], HA_DEV_ONOFF_SWITCH);
  }
  for (int index = 0; index < relay_clusters_cnt; index++)
  {
    zigbee_endpoint_init(&endpoints[switch_clusters_cnt + index], HA_DEV_ONOFF_OUTPUT);
  }

  basic_cluster_populate(&basic_cluster);
zigbee_endpoint_add_cluster(&endpoints[0], 0, ZCL_CLUSTER_OTA);

for (int index = 0; index < switch_clusters_cnt; index++)
{
switch_cluster_add_to_endpoint(&switch_clusters[index], &endpoints[index]);
}
for (int index = 0; index < relay_clusters_cnt; index++)
{
    relay_clusters[index].scene_cluster = scene_clusters + index;
    scene_clusters[index].relay_cluster = relay_clusters + index;

```

marazmarci (changes requested):

marazmarci:
The On With Timed Off changes will need to be rewritten because in the HAL PR #178, the polling-based logic is replaced with a callback-based approach, and the logic in this PR relies on polling for counting the remaining time.

Probably the way to go would be to use "tasks", but I don't yet know how they work in the Telink SDK, so I cannot provide more details now.

ThatMishakov:
> The On With Timed Off changes will need to be rewritten because in the HAL PR #178, the polling-based logic is replaced with a callback-based approach, and the logic in this PR relies on polling for counting the remaining time.
> 
> Probably the way to go would be to use "tasks", but I don't yet know how they work in the Telink SDK, so I cannot provide more details now.

I'll take a look at it, the last time I've looked at the timers it didn't look very difficult with the Telink SDK, although I'm not sure how is the Silabs' SDK

Also, I'll fix the issue with the clusters, I think I've picked the wrong commit when rebasing since it worked well in my branch

marazmarci:
> I'll take a look at it, the last time I've looked at the timers it didn't look very difficult with the Telink SDK, although I'm not sure how is the Silabs' SDK

Pro tip: check out `hal_tasks_schedule` on the `hal` branch.

PR #174, opened 2025-10-16
Add scenes out cluster (+ add some mandatory attributes while at it)

ThatMishakov:
This pull request adds a double click, scene out cluster, 4 new attributes for it (in OnOff switch config cluster), and also adds mandatory Cluster revision to all clusters in switch_cluster.c

This allows to cycle through scenes with a double click of the switch, which has many (awesome) applications with bindings, which include the following:
- Controlling smart lights' colors/color temperature/brightness
- Controlling multiple lights or relays (e.g. turning specific lights on and off)
- Turning all lights on or off (e.g. when leaving a room)
- Being able to bind newly supported scene switches to ... scenes
- Automations based on double clicks and next scene attribute value

The following new attributes are added:
- Next scene: next scene to be applied with a double click
- Scene count: number of scenes to cycle through. 0 disables the double click (and the delay associated with its detection)
- Scene start offset: the starting scene
- Scene group ID: group ID to be sent in the recall scene command. Also, the idea was to deduce the group id based on the binding if this is set to 0xffff, but more on that below
- Scene recall time (currently broken?): recall time to be sent in the recall scene command
- Multistate attribute gets 2 new values, MULTISTATE_DOUBLE_PRESS and MULTISTATE_SINGLE_RELEASE

For example, if scene count is set to 3, and scene offset is set to 10, double clicks will recall scenes 10 -> 11 -> 12 -> 10 and so on.

Some considerations/TODO stuff:
- Double click detection adds a little bit of delay — I've hardcoded it to 250 ms — to the `short_press` actions, which is kinda against the spirit of the project, but I think it's an acceptable compromise for the (optional) extra functionality
- The current 50ms debounce window is kinda very large for my liking, and misses double presses if you do it too fast (I've tried 10ms, and it worked much better with my relays and momentary switches). Perhaps, it could be made configurable or just reduced if there are no problems with touch sensors and so on.
- This adds a lot of attributes, and I've already heard people complain that there are too many of them. While I personally like having all the possible functionality, perhaps some of them could be removed and hardcoded to some sensible values, or there can be multiple versions of converters or something like this.
- Since this adds a lot of settings, it might need a configuration version bump? Otherwise, the switch configuration gets kinda broken after the update.
- I'm not very familiar with the SDK, and couldn't get the group ID deduction to work based on the bindings (but I didn't put much effort into it). Tbh, maybe this could be removed.
- The `button.c` is a bit of a mess, and by the time @andrei-lazarov told me about https://github.com/romasku/tuya-zigbee-switch/pull/160#issuecomment-3395205106 it was too late. But it works
- Something needs to be done about multiple presses to reset. If you cycle through 5 scenes too quickly, your relay gets reset with how things are at the moment
- So far I've only changed the custom converter for TS0002-OXT-CUS (my relays), and also Zigbee2mqtt complains when reading 0xffff from the attributes. This needs to be fixed and applied to all the converters, or the new ones (I haven't really paid attention to them), so I don't know what's the best course of action
- So far I have only tested it with my relay and momentary NO switches, it might be broken on other devices

### DIFF
--- a/zigbee2mqtt/converters/switch_custom.js
+++ b/zigbee2mqtt/converters/switch_custom.js
@@ -88,7 +88,7 @@ const romasku = {
             name,
             endpointName,
             access: "STATE_GET",
-            lookup: { released: 0, press: 1, long_press: 2, position_on: 3, position_off: 4 },
+            lookup: { released: 0, single_press: 1, long_press: 2, position_on: 3, position_off: 4, double_press: 5, single_release: 6 },
             cluster: "genMultistateInput",
             attribute: "presentValue",
             description: "Action of the switch: 'released' or 'press' or 'long_press'",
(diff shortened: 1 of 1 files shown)


andrei-lazarov:
Thank you for implementing scenes! This is very cool and meaningful for the project 😃👍🏻

Right now the PR is a little overwhelming for me.. I can't review it because I have never used scenes.  
I'll have to play around and test to give some feedback.

Anyway I suggest we fix the TODO step-by-step before we merge scenes into main.  
Let's tackle these separately:

- [ ] 2 converters: simple and advanced
- [ ] refactor `button.c`
- [ ] configurable de-bounce
- [ ] configurable reset by spamming
- [ ] configurable (and optional) double-click

I can do the tasks (I wanted to do them anyway), but it will take some time.  
My motivation comes and goes randomly 🙂 

Right now I want to fix the generic converters and test them with older Z2M versions.. see how far we can go until they break. Then split them into 2 versions.

ThatMishakov:
>     * [ ]  configurable (and optional) double-click

The double click should already get disabled if you set the scene_count to 0. It can be checked by locking at `press_action`, if it's disabled, you will see `single_press` then `released` when clicking a button, if it's not, there should be an extra `single_release` state, I think this might be enough?

Also, the main problem I see with configurable de-bounce (and reset) is that that would be even more attributes, if it's not an issue I guess I can also do it myself

PR #180, opened 2025-10-27
Change `APP_RELEASE` for hal branch

clumsy-stefan:
- Change `APP_RELEASE` in version_cfg.h to identify this branch uniquely
- allow upgrades without `-FORCE`
- prevent accidental downgrades


romasku:
Thank you!

PR #185, opened 2025-11-02
Fix switch action and indicator reporting in the Z2M converter (HAL)

marazmarci:
When testing the HAL changes (#178), I found that the switch cluster's press action and the relay cluster's indicator state are not reported. However, when reading the multistate input's `presentValue` (switch action) or the onOff cluster's `0xFF02` attribute (indicator state) from the dev console in Z2M, I get the correct values.

The solution is to configure reporting for the attributes from the converter/quirk.

**I only implemented it for the Z2M converter, and I didn't even try fixing the ZHA quirk, because I'm only using Z2M and I don't have an easy way to set up a test env with ZHA. Also, I only tested this on a Telink device, and I don't know if the SiLabs version needs reporting to be configured or not, as I don't (yet) have a SiLabs test device.**

As far as I understand, this issue was caused by the following change:
https://github.com/romasku/tuya-zigbee-switch/blob/4b69826aa28aec343300dc73028c7ac1afedf872/src/telink/hal/zigbee_zcl.c#L167-L170

I support the change to rely on reporting logic built-in to the SDK and don't just eagerly push changed attribute values from the device, but with this, we also need to configure reporting from the converter/quirk.

**While my changes fix both problems (switch action, indicator state), there's a small inconvenience:
The indicator state is using the 0xFF02 custom attribute, which is not known by Z2M, and the (new Windfront) web displays it by selecting the first attribute in the list, which happens to be `elkoOnTimeReload` (which has nothing to do with 0xFF02, as it doesn't even have the same attribute ID), and because of this, the user can't manually disable or modify reporting for the indicator state attributes. But the reporting works, because the reporting was configured for the correct attribute ID from the converter.**

### Other things
In 3e9560cc I added a test case which I believe has added value over the existing `test_momentary_mode_multistate_value` test case because that one only test the attribute's value via explicitly reading it. My new test checks if the switch action is reported or not. If you don't think it's useful, feel free to leave it out.

I noticed that the relay onOff cluster's onOff state was configured to report every minute, and while I was there in the converter, I changed it to `repInterval.MAX` in 3f707dbb, because I don't think reporting this often is required, and we could spare some unnecessary Zigbee traffic.

### DIFF
--- a/helper_scripts/templates/switch_custom.js.jinja
+++ b/helper_scripts/templates/switch_custom.js.jinja
@@ -233,7 +233,7 @@ const definitions = [
             const endpoint{{loop.index  + (device.switchNames | length)}} = device.getEndpoint({{loop.index  + (device.switchNames | length)}});
             await reporting.onOff(endpoint{{loop.index  + (device.switchNames | length)}}, {
                 min: 0,
-                max: constants.repInterval.MINUTE,
+                max: constants.repInterval.MAX,
                 change: 1,
             });
             // indicator:


romasku:
Great job, thank you very much!

andrei-lazarov:
This PR fixed reporting for switches with state-indicator-LEDs, but broke reporting for switches without them.

Z2M throws the unsupported attribute error when they are missing, so we should only add the reporting for devices that need it.

marazmarci:
Ohh, sorry, good catch! I surely need more than one kind of test device :smile:

romasku:
Thank you for spotting this. It should be fixed by https://github.com/romasku/tuya-zigbee-switch/pull/178/commits/e979ccbfa8476809e10d473a72716d7f4c203470

PR #196, opened 2025-11-18
Fix BiStable relays

aceindy:
Replaces #131 by @rastakozza

As suggested, it now work as this:
In both relay_on, relay_off clear both pins first.
Add scheduled_off_task field to relay_t, and if it is not NULL cancel it first
Turn on the proper pin (pin or off_pin).
Schedule task to clear and store it as relay->scheduled_off_task = hal_tasks_schedule

### DIFF
--- a/src/base_components/relay.c
+++ b/src/base_components/relay.c
@@ -42,11 +42,6 @@ void relay_init(relay_t *relay)
   // Mark that no task is active
   relay->clear_task_active = 0;
   
-  // Prepare the task structure *once*
-  relay->clear_task.handler = relay_clear_handler;
-  relay->clear_task.arg = relay;
-  hal_tasks_init(&relay->clear_task);
-
   //switch off relay
   relay_off(relay);
 }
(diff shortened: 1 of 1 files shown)


aceindy:
@romasku Does this looks better?
Just doing some final testing, then I will make it ready for review 😉 

We probably need some changes in the tests though, as it currently doesn't know how to deal with BiStable relays 😄 
```
=========================== short test summary info ============================
FAILED tests/test_base_components.py::test_relay_off_pin_mutual_exclusion - AssertionError: assert False is True
 +  where False = <bound method Device.get_gpio of <tests.conftest.Device object at 0x7f13949832c0>>('C1')
 +    where <bound method Device.get_gpio of <tests.conftest.Device object at 0x7f13949832c0>> = <tests.conftest.Device object at 0x7f13949832c0>.get_gpio
```

aceindy:
Looks like it's working fine (but test need to be updated, which is out of my scope)

PR #214, opened 2025-11-28
Add support for ZG-807Z with external PA

Rococo88:
mac_phy.c changes are placed it sdk_patch directory however I am not sure how this will be included in local setup with make_install.sh, needs further investigation, device_db.yaml adds partially spoofed device as placing device name "CK-BL702-ROUTER-01(7018)" recognized by Z2M in config string directly is not parsed properly when building firmware.


thecampagnards:
Hello ! Does make it works without tuya router ? Thx !

Rococo88:
Well, it worked in my Z2M network without any tuya firmware in the coordinator as a signal repeater/router.

PR #222, opened 2025-12-01
Add support for Novato ZNR01

Flee-Time:
Added support for the Novato ZNR01 no-neutral single-gang relay module.

[Store Link.](https://novato.com.tr/urun/tuya-mini-zigbee-notrsuz-tek-cikisli-role/)

- **Hardware:** Novato ZNR01 (Tuya White Label)
- **MCU:** Telink TLSR8258 (ZT5 Module)
- **Pinout:**
   - Button: B7
   - LED: B1
   - Relay/Triac: B5
   - External Switch: C2
- **Status:** Fully supported and working.
- **Manufacturer ID:** `_TZ3000_30jqysvd`

The factory firmware is missing the ota cluster so the first flash needs to be done by wire. There's no coil whine because this module uses a triac to control the current flow of the attached light/device. Need to remove two capacitors to be able to solder the flashing wires. Seems to be fully working with no issues.

IMG20251201200051
IMG20251201200105
IMG20251201200136
IMG20251201200013


Flee-Time:
btw, in the meanwhile i've been trying to get stock-to-custom OTA working. the device advertises the OTA cluster when interviewed, but it completely fails to answer requests for currentFileVersion or imageType (even when waking it up).

couldnt get it working sadly, would be much better to convert without a flasher. im leaving the tuya manufacturer id and image type blank just so people dont try to flash via ota and end up with a possible brick.

Flee-Time:
just analyzed the stock firmware backup to be 100% sure.

even though the device lists the OTA cluster (0x0019) in its descriptor, the actual firmware header (at offset 0x8000) is missing the standard Tuya ID and Image Type definitions required for OTA. It looks like they copy-pasted the cluster descriptor but stripped out the actual OTA stack.

so stock-to-custom OTA is definitely impossible on this unit.

andrei-lazarov:
I see you got accustomed quickly 🙂
If you have the stock firmware backup, please upload it to `bin/_factory/`

Flee-Time:
yep will do in a bit, let me write the backup to the device and reset the pairing, the backup currently has my network keys.

andrei-lazarov:
Oh, don't go through the soldering hassle again. It's ok - we don't really do anything with the backups.

Flee-Time:
too late :D
dont worry about it, taking the backup again now

i would have just zeroed the nvram section of the backup but dont want to risk it not being bootable if someone needs it.

Flee-Time:
should be good to go now.

andrei-lazarov (approved):

romasku:

PR #223, opened 2025-12-02
Add support for Novato ZRM01

Flee-Time:
# Added support for the Novato ZRM01 single-gang relay module (Neutral Required).

[Store link.](https://novato.com.tr/urun/tuya-mini-zigbee-tek-cikisli-akilli-role/) (Seems to be selling a different revision/model now)

- **Hardware:** Novato ZRM01 (Tuya White Label)
- **MCU:** Telink TLSR8258 (ZTLC9 Module, see below)
- **Pinout:**
    - Button: B4
    - LED: C2
    - Relay: D2
    - External Switch: C4 (ADC pad on module, see below)
    - RF Header: C3 (Not supported, see below)
- **Manufacturer Model:** `_TZ3000_hdc8bbha`
- **Z2M Model:** `QS-Zigbee-SEC01-U`
- **Status:** Fully supported and working (RF daughterboard ignored).

# Notes:
- **Module:** The `ZTLC9` Zigbee module seems to be a form-factor clone of the `WBLC9` but uses a Telink TLSR8258 chip instead.
- **Switch Input:** The `ADC` pad exposed on the module is connected to pin `C4` of the TLSR8258. I have configured the switch input as floating because there is a physical 10k pull-up resistor on the board itself.
- **RF Header:** The device has a header for a 433MHz daughterboard connected to pin C3. Since the custom firmware does not support RF decoding, this pin is left unused. However, the header is accessible through the case, so it could potentially be repurposed by advanced users.
- **OTA:** Similar to the other Novato module, the stock firmware is missing the standard OTA `imageType` and headers. Tuya-to-Custom OTA is impossible on this device; it must be flashed via wire for the initial install.
- **Hardware:** There is no audible coil whine from the relay when it is on or off.
---
IMG20251202130157
IMG20251202122759
IMG20251202122806
IMG20251202124211
IMG20251202124114


Flee-Time:
Yep, all good.

andrei-lazarov:
Also - forgot to say - it would be nice if you could force-push all your changes into a single commit.
But it's good to merge anyway 👍🏻

Flee-Time:
Ah, will do for the next device i port. Been using the github web edit, havent cloned the repo locally.

Flee-Time:
Also, another question. In zigbee2mqtt the zigbee model for this one is TS000F. But the 2 gang version of this has the same zigbee model (TS000F). Only the manufacturer model is different. how should we save this in the yaml file so they don't conflict?

andrei-lazarov:
Ah, I missed that, and wrongly put it as TS0001. They can be: NOVATO_TS000F_1 and NOVATO_TS000F_2.

andrei-lazarov (approved):

romasku:
Thank you!

PR #232, opened 2025-12-06
Add support for ZBMINIL2

romasku:
Check [guide](https://github.com/romasku/tuya-zigbee-switch/blob/main/docs/flashing/zbminil2.md) for details.


philipgo:
Has anyone tried whether this works for the ZBMINIR2 Extreme as well?

andrei-lazarov:
Hi @philipgo! We can port the firmware for the ZBMINIR2 as well. I have a few.
Are you having issues with the stock firmware / what features do you need?
It received bindings in the v1.0.8 update from SONOFF, so it's pretty good now.

philipgo:
That would certainly be interesting. I find them to be very good switches, especially with the radio turbo mode enabled, but I do not like the group binding implementation.

andrei-lazarov:
I'll let you know when it's available

johnmph:
Hello, I am interested also for the ZBMINI2R, I want to avoid or change the way the factory reset works with the external input, and I can do that with your firmware (MULTI_PRESS_CNT_TO_RESET).
Thank you

pslavov:
s it possible to flash the firmware using USB UART controller ?

andrei-lazarov:
@pslavov No, you need the J-link v9 programmer

robin0793:
Hi, I just flashed my ZBMINIL2 using a J-Link V9, I followed the flashing instructions and everything went well, but the device seems dead. After flashing the firmware, the LED did not start to blink, nor does it connect to the zigbee network. Button press doesn't do anything either.

Chip type: EFR32MG22C224F512GN32 (I selected EFR32MG22CxxxF512 in Simplicity Commander)

I flashed efr32mg22_bootloader_module_sonoff_zbminil2.s37 and tlc_switch-1.1.2-9a4eb422.s37 (end device)

andrei-lazarov:
@robin0793 there are at least 2 variants of the ZBMINIL2, with different flash chips.

The firmware from the main branch will only work on the older ZBMINIL2 that romasku has.
It's very likely you have the new variant.

I don't remember what's the status. There is some progress here:
- https://github.com/romasku/tuya-zigbee-switch/pull/294

And a discussion on Discord with some test builds. Can you join and leave a message? (link in readme)

PR #256, opened 2025-12-19
Ts0002 tz3000 mufwv0ry

9mrcookie9:
Adding support for PSMART 2 Gang switch. 
PCB same as psmart 1. 
Alt configuration allows you to use green (network) led manually.
Third relay is used to expose backlight.


romasku:
Thank you!

andrei-lazarov:
Fixed converter and removed double pin assignment in #260

PR #258, opened 2025-12-19
Added psmart switches stock firmware

9mrcookie9:
Hi, 
I was not aware of this section of the repository. 
Both of the files were generated via shell comand from instructions. 
I renamed files with repository based name for both of them.


romasku:
Thank you!

PR #265, opened 2025-12-24
Fix: Add missing --sdk flag to slc signature trust command

bobmoroshev:
The second slc signature trust command in tools.mk was missing the --sdk flag, causing the build to fail with 'No SDK' error. and then
> Clear out index files
> Process completed with exit code 1.


andrei-lazarov:
@hurrian apply this fix to finish your build

hurrian:
Works for me, thanks!

romasku:
Thank you, sorry for the delay.

PR #260, opened 2025-12-21
🗃️ Update database

andrei-lazarov:
See each commit for info ⬆

PR #269, opened 2025-12-29
Add support for _TZ3000_kea5qgnd (MANHOT 1g, Telink TLSR8258)

hurrian:
Single gang version of _TZ3000_tqwydnqn (#128)

### DIFF
--- a/device_db.yaml
+++ b/device_db.yaml
@@ -1515,6 +1515,33 @@ MANHOT_SWITCH_1_TS0013:
   store:
   - https://www.aliexpress.com/item/1005008300010160.html
   - https://www.aliexpress.com/item/1005008867890261.html
+MANHOT_SWITCH_A_TS0011:
+  human_name: Manhot 1-gang switch 🅰
+  category: switch
+  power: mains
+  neutral: optional
+  device_type: end_device
+  tuya_model_name: TS0011
+  tuya_manufacturer_name: _TZ3000_kea5qgnd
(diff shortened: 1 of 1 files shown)


andrei-lazarov:
Thank you for adding your switch!
I assume you tested the device and it works well, so we could also call it 'fully_supported'?

Can you target [my branch](https://github.com/romasku/tuya-zigbee-switch/pull/260) (latest db) instead of romasku/main, to avoid conflicts?
I can also add it myself if you agree.

andrei-lazarov:
Forgot to say, please also change the database entry with fw_image_type 43599.

Then, to fix the already-flashed devices, append `i43599;` to their config string.

hurrian:
> Forgot to say, please also change the database entry with fw_image_type 43599.
> 
> Then, to fix the already-flashed devices, append `i43599;` to their config string.

Rebased, thanks! The commit in this PR follows the schema in romasku/main; my _TZ3000_kea5qgnd_build branch is rebased on top of yours.

andrei-lazarov:
Thank you! You may close this PR.
I added your commit to mine.

hurrian:
@andrei-lazarov By the way I noticed you templated support for the 2 gang version of this device. The ID is `_TZ3000_toaaawnr` and the pinout is correct.

andrei-lazarov:
Thanks. Added and started the build action on db/wip_build if you need it

Issue #270, opened 2025-12-29
Add Girier TS130F Dual Shutter Module


LorandBiro:
As we briefly discussed in https://github.com/romasku/tuya-zigbee-switch/discussions/268 already, I would like to explore what it would take to add shutter functionality to this firmware. I have single- and dual-channel Girier shutter modules in my home, and they already provide basic shutter control reasonably well. They can be calibrated, and the motor direction can be swapped in case they are wired incorrectly. However, there are a few areas I would like to improve:

- The roller shutters I know and have used have significant dead zones at the upper and lower ends of travel. For example, my roller shutters take around 20 seconds to go from one end position to the other, but for the first 3-4 seconds there is no visible movement. Then they start moving, and after they are fully closed, the motor can still be heard running for a few more seconds. These two dead zones mess up position calculation. If I set the roller shutter from 0% to 10%, I hear the motor running for 2 seconds, but the shutter does not actually move at all. When setting a specific position, I often do not have an intuitive understanding of what percentage corresponds to the desired position. A better firmware would allow calibration of these two dead zones so the position value better matches the actual roller blind position.
- I have [Velan Life Roller blind push-buttons](https://www.legrand.com/ecatalogue/en/catalog/products/roller-blind-push-button-valena-life-6-250-v-with-cover-plate-white-752130?category_id=64231). These have a stop button in the middle that effectively closes both direction switches at the same time. A better firmware would interpret simultaneous switch presses as a stop command.
- My 1-gang and 2-gang Girier shutter modules behave inconsistently. One can be stopped by momentarily pressing the opposite direction, while the other actually starts moving the motor in the opposite direction. One can be controlled by holding down the button, so it only moves while the button is pressed, while the other does not support this. A better firmware would be more consistent and maybe even allow these behaviors to be configured to match the home owner’s preference.

# What I did

I figured out the port mapping, created a build on my fork, successfully OTA-updated the original firmware with the custom firmware on a dual-channel Girier shutter module, and verified the port mapping through the config string.

There is one quirk on this device: the on-board reset button appears to be wired to the same pin as the first switch input. I have not fully explored the PCB yet, so I do not know why another pin was not used. On this device, pressing the reset button is equivalent to pressing the first switch.

# Next steps

[...]

Issue #263, opened 2025-12-23
Improving z2m integration


Koenkk:
Hi!

First of all; I've been using this project for a couple of weeks now and I'm really happy with it 😄 

Given that the initial setup can be a bit complicated, I want to discuss some options to make this easier such that more users can use this:

- The [customer converters](https://github.com/romasku/tuya-zigbee-switch/blob/main/docs/updating.md#quirks--converters) can become part of https://github.com/koenkk/zigbee-herdsman-converters such that devices running this firmware are supported out-of-the-box (bonus: this will automatically add it to https://www.zigbee2mqtt.io/supported-devices/)
- For the initial OTA flashing (currently requires [local OTA index](https://github.com/romasku/tuya-zigbee-switch/blob/main/docs/updating.md#ota-index) + customer converters with `ota: true`): we can use the new [actions](https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/converters/actions.ts) API (recently introduced by @nerivec). My initial idea is that we add a custom actions for this project which pulls the OTA from this repo and starts the OTA process with the selected device. This action can then be triggered via either the frontend or MQTT.

Let me know what you think!

andrei-lazarov:
Hi Koenkk, thanks again for your interest, and Happy Holidays! 😃  

**Full Z2M integration is definitely the goal,** I am very excited about that!  
- I didn't push in this direction, because I see it equivalent to a 'public release', which the project isn't quite ready for
_(We can barely handle more support and feature requests, and we have a long to-do list)_.
- Ultimately, Romasku will decide how to proceed.

**We should unify and standardize the firmware, before adding OTA and converters to Z2M** _(see to-do list)_.

**The biggest issue right now is that we build an image for every device.** 
Firmware is identical, only the default GPIO mapping is different. 

- @romasku I propose we build it only once, and **create a script to hex-edit the config string** _(on-demand, to skip storing multiple binaries)_. 

- Then I think we could **use Nerivec's Z2M actions to execute that script** and grab device info, to automatically inject the correct pinout and Tuya imageType needed for initial OTA.
 If device is not in db, allow flashing the firmware with blank GPIO _(user sets it afterwards)_

<details><summary><b>Project to-do list</b></summary>

- [ ] **Unify**
  - [ ] Generic firmware image 
  - [ ] Generic converters
  - [ ] Mixed index _(router + end\_device)_
- [ ] **Standardize**
[...]

PR #271, opened 2025-12-29
Add Girier TS130F Dual Shutter Module

LorandBiro:
As discussed in https://github.com/romasku/tuya-zigbee-switch/discussions/268 and https://github.com/romasku/tuya-zigbee-switch/issues/270, this PR adds the Girier TS130F dual shutter module to `device_db.yaml` as a first step toward adding shutter support to the firmware. At the moment, the device is added simply as a 4-gang switch module, so installing this firmware will remove the shutter functionality from the device.

Regarding the device database, I have two questions:

- Are there any guidelines for choosing a new model name in the config string? I tried to find a pattern but did not see a consistent one, and this seemed like the best fit.
- I found the largest existing `firmware_image_type` value and incremented it by one. Is that the correct approach here?

I also made the following related changes and fixes in this PR:

- The single-channel shutter module is defined in `tuya.js`, but the dual-channel shutter module is defined in [lonsonho.ts](https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/lonsonho.ts) in the `zigbee-herdsman-converters` repository. Because of this, I extended the `make_z2m_tuya_converters.py` script to include the Lonsonho manufacturer. Is this the right approach?
- In the `tuya_with_ota.js.jinja` template, I fixed two pieces of outdated or inaccurate information.
- In `tools.mk`, I had to provide the `--sdk` parameter to avoid build failures: https://github.com/LorandBiro/tuya-zigbee-switch/actions/runs/20579367295/job/59103410520#step:8:77

### DIFF
--- a/device_db.yaml
+++ b/device_db.yaml
@@ -3138,3 +3138,29 @@ SWITCH_ZEMISMART_TS0601_6GANG:
   info: Needs pinout. Secondary MCU. 6-gang!
   threads: https://github.com/romasku/tuya-zigbee-switch/issues/209
   store: https://www.zemismart.com/products/tb26-6
+MODULE_GIRIER_TS130F_2GANG:
+  human_name: Girier TS130F Dual
+  category: module
+  power: mains
+  neutral: required
+  device_type: router
+  tuya_model_name: TS130F
+  tuya_manufacturer_name: _TZ3000_j1xl73iw
+  stock_converter_manufacturer: Lonsonho
+  stock_converter_model: TS130F_dual
+  override_z2m_device: TS130F_GIRIER_DUAL
+  tuya_module: ZTU
+  mcu_family: Telink
+  mcu: TLSR8258
+  config_str: j1xl73iw;TS130F-GIR-DUAL;LC1;SB4u;RC0;SD2u;RC4;SC3u;RD4;SC2u;RD7;
+  alt_config_str: null
+  old_manufacturer_names: null
+  old_zb_models: null
+  tuya_manufacturer_id: 4417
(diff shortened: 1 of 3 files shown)


andrei-lazarov:
Thanks for adding the device! 🙂 Very cool that any module can be a shutter module.

Firstly, please also track my fork, or check PRs before submitting, to avoid conflicts. 
**Target andrei-lazarov/db/wip with your pull request** - The latest DB is usually here. 
- https://github.com/romasku/tuya-zigbee-switch/pull/260

The project needs some unification and standardization, as I explained in this issue with a _brief_ to-do list:
- https://github.com/romasku/tuya-zigbee-switch/issues/263
- I managed to standardize KEYS, so **please call it MODULE_GIRIER_TS130F_2GANG**
- **Model name is good**, no pattern _yet_
  - Preferably unique, while keeping config_string under ~74 chars (Zigbee message becomes too long) 
  - My idea is to make the manufacturer_name 'romasku', and model_name its own field, longer and more descriptive
- **Fw image type is good**, no pattern _yet_
  - Preferably unique currently 
  - However I see no point in unique numbers, as the firmware is identical for all devices.. will probably be standard in the future
- **Tuya converters are correct.** 
  - But we will probably drop them, in favor of Z2M actions, as discussed in issue 263 
- **Build fix is correct**, was already added to main

Also please send me a private message / mail

LorandBiro:
Thanks for the review! I had to recreate the PR on your fork so I could target `andrei-lazarov/db/wip`.
Sorry for the messy commit history! I should have closed the PR before rebasing.

PR #272, opened 2025-12-30
Align pairing mode trigger with documented behavior (10 presses)

LorandBiro:
When following the documentation to enter pairing mode by pressing the switch 10 times, I found that it actually requires 11 presses. I believe this may be an unintentional off-by-one error.

The constant `MULTI_PRESS_CNT_TO_RESET` is defined as `10`, and the documentation mentions 10 presses, but the condition uses `>` (greater than) instead of `>=`:

If requiring 11 presses was intentional, please let me know and I can update the documentation instead. Happy to adjust this PR based on the intended behavior!


romasku:
Thank you!

PR #277, opened 2026-01-06
Align leave-network unit test with new press count

LorandBiro:
PR #272 changed the leave-network press count from 11 to 10, but I didn’t realize there was a unit test covering this. This PR updates the unit test to verify the exact number of presses.


romasku:
Thank you for spotting this!

PR #278, opened 2026-01-06
Setting Home Assistant entity categories for cleaner default dashboard

LorandBiro:
This PR is a proposal to improve the default Home Assistant experience for devices using this firmware. The goal is to reduce dashboard clutter by adjusting how entities are categorized during discovery.

When a device running this firmware is discovered in Home Assistant via MQTT and Zigbee2MQTT, it creates many entities, most of which are categorized as Control or Sensor entities. Home Assistant automatically shows these on the default dashboard, which can be inconvenient if the dashboard is used for day-to-day control rather than device management.

This is how the device currently appears on the Home Assistant device page and default dashboard:

image
image

Users need to manually hide or disable each entity if they want to keep the dashboard clean. After explicitly setting the `entityCategory` fields, the device page and default dashboard for a newly added device look like this instead:

image
image

A note on the sensor vs diagnostic categorization. While the press action could reasonably be exposed as sensors, they are not the primary purpose of the device, and most users likely do not want to see them on the dashboard by default. Since sensor entities are shown automatically, marking these as diagnostic felt like a better default to reduce clutter.

This is only a suggested change and I am happy to adjust or close the PR if there is a different preferred approach.

### DIFF
--- a/helper_scripts/templates/switch_custom.js.jinja
+++ b/helper_scripts/templates/switch_custom.js.jinja
@@ -38,6 +38,7 @@ const romasku = {
             - toggle_simple: Any press of physical switch will TOGGLE the relay and send TOGGLE command to binds
             - toggle_smart_sync: Any press of physical switch will TOGGLE the relay and send corresponding ON/OFF command to keep binds in sync with relay
             - toggle_smart_opposite: Any press of physical switch: TOGGLE the relay and send corresponding ON/OFF command to keep binds in the state opposite to the relay`,
+            entityCategory: "config",
         }),
     switchMode: (name, endpointName) =>
         enumLookup({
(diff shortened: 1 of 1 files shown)


LorandBiro:
@andrei-lazarov 

> Thanks! This was long overdue 🙂 We'll just have to repeat it for the generic converter

Is that the [rs/generic-converter](https://github.com/romasku/tuya-zigbee-switch/tree/rs/generic-converter) branch? I'm happy to open a PR for that branch too.

andrei-lazarov:
Yes

LorandBiro:
Got it. I applied the same changes to the generic converter. Here's the other PR: #279

romasku:
Thank you!

PR #279, opened 2026-01-06
Set Home Assistant entity categories in generic converter

LorandBiro:
This is a follow-up to #278.

This PR sets the category field on all expose objects. The goal is to reduce Home Assistant dashboard clutter by adjusting how entities are categorized during discovery. I am applying the same changes here to keep behavior consistent across the two converters.

There is a strange behavior around the device config string in this converter. If I do not set a category, it is displayed as a sensor in Home Assistant. If I try to set it to **diagnostic**, the converter fails to load because writable exposes cannot be set to **diagnostic**:

```
[2026-01-06 22:56:19] error: 	zhc: Failed to process exposes for '0xa4c1381fd6ddd684' (AssertionError [ERR_ASSERTION]: Diagnostic expose must not be settable
    at Composite.validateCategory (/app/node_modules/.pnpm/zigbee-herdsman-converters@25.83.1/node_modules/zigbee-herdsman-converters/src/lib/exposes.ts:72:23)
    at Composite.withCategory (/app/node_modules/.pnpm/zigbee-herdsman-converters@25.83.1/node_modules/zigbee-herdsman-converters/src/lib/exposes.ts:62:14)
```

If I set it to **config**, Zigbee2MQTT decides this does not make sense and overrides it to **diagnostic**: https://github.com/Koenkk/zigbee2mqtt/blob/master/lib/extension/homeassistant.ts#L1241

I have not made further changes to the composite wrapper around the device config and left its current behavior unchanged. With these changes, the Home Assistant device page looks like this:

image

### DIFF
--- a/helper_scripts/templates/switch_custom.mjs.jinja
+++ b/helper_scripts/templates/switch_custom.mjs.jinja
@@ -62,7 +62,8 @@ const customExposes = {
       .enum(`switch_mode`, ea.ALL, Object.keys(SWITCH_MODE_ENUM))
       .withLabel(`Switch ${switchIndex} mode`)
       .withDescription("Select the type of switch connected to the device")
-      .withEndpoint(`switch_${switchIndex}`),
+      .withEndpoint(`switch_${switchIndex}`)
+      .withCategory("config"),
   switchActions: (switchIndex) =>
     e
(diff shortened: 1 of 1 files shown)


romasku:
Thank you for porting this to generic converter as well!

PR #283, opened 2026-01-08
🗃️  Update database

andrei-lazarov:


bobmoroshev:
Fixed support of Moes ZM4LT2
https://github.com/andrei-lazarov/tuya-zigbee-switch/pull/10

PR #288, opened 2026-01-13
Basic cover cluster implementation

LorandBiro:
I initially created a prototype on my fork with an almost complete feature set, but the resulting change set turned out to be too large to review comfortably. To keep things manageable, this PR includes the smallest reasonable subset of the window covering functionality.

At the moment, it only implements the Open, Close, and Stop commands with the required safety delays. The remaining functionality could be completed in one or two follow-up PRs, which I think would make the review process much easier overall.

I tried to keep the changes as isolated as possible to avoid increasing the complexity of the existing codebase. That said, a few updates to existing components were necessary:

* HAL: The `cluster_id` parameter type was changed from `uint8_t` to `uint16_t` to allow the window covering cluster ID to work correctly.
* Converter: The relay and switch endpoints are now optional, so the code dynamically finds the first suitable endpoint when adding the config string and the indicator LED.

I also noticed the `uncrustify.cfg` file in the repository and applied it to the new code. However, it seems that much of the existing codebase does not strictly follow this formatting. Please let me know if this is acceptable, or if you would prefer that I align the new code more closely with the existing style.

Feedback is very welcome, and I am happy to adjust the approach if needed.

### DIFF
--- a/helper_scripts/templates/switch_custom.js.jinja
+++ b/helper_scripts/templates/switch_custom.js.jinja
@@ -318,7 +318,6 @@ const definitions = [
                 },
             ]);
             {% endfor %}
-
             {% for coverName in device.coverNames %}
             const cover{{loop.index}} = device.getEndpoint({{loop.index + (device.switchNames | length) + (device.relayNames | length)}});
             await reporting.bind(cover{{loop.index}}, coordinatorEndpoint, ["closuresWindowCovering"]);
--- a/helper_scripts/templates/zha_quirk.py.jinja
+++ b/helper_scripts/templates/zha_quirk.py.jinja
@@ -4,6 +4,7 @@ from zhaquirks import CustomCluster
 from zigpy.quirks.v2 import QuirkBuilder, ReportingConfig, SensorDeviceClass, EntityType
 from zigpy.zcl import ClusterType, foundation
 from zigpy.zcl.clusters.general import OnOffConfiguration, SwitchType, MultistateInput, OnOff, Basic
+from zigpy.zcl.clusters.closures import WindowCovering
 from zigpy.zcl.foundation import ZCLAttributeDef
 import zigpy.types as t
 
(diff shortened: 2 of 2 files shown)


device_db.yaml
```diff
@@ -766,7 +766,7 @@ MODULE_GIRIER_TS130F_2GANG:
   tuya_module: ZTU
   mcu_family: Telink
   mcu: TLSR8258
-  config_str: j1xl73iw;TS130F-GIR-DUAL;LC1;SB4u;RC0;SD2u;RC4;SC3u;RD4;SC2u;RD7;
+  config_str: j1xl73iw;TS130F-GIR-DUAL;LC1;CC0C4;CD4D7;
   alt_config_str: null
```
andrei-lazarov: Let's keep the switch config as alt_config

```suggestion
  alt_config_str: j1xl73iw;TS130F-GIR-DUAL;LC1;SB4u;RC0;SD2u;RC4;SC3u;RD4;SC2u;RD7;
```

romasku:
Thank you for this contribution! I didn’t have time for a proper review yet, but at first glance it looks very nice. I’ll try to do a more thorough review soon.

Regarding formatting -- that’s a bit of technical debt on my side. I haven’t settled on a style I really like yet, so the formatting is somewhat inconsistent. I’ll probably reformat all the code later, so don’t worry too much about it -- your code looks more than clean enough to be merged.

Could you also add information about this new config entry to the documentation, specifically to [this file](https://github.com/romasku/tuya-zigbee-switch/blob/main/docs/contribute/porting.md)?

LorandBiro:
Sure, I added a new line to the `porting.md` file to document the new config entry. In the next PR, I plan to add cover switches, and then it will make sense to include an example as well.

Would you like me to indicate that this is a work-in-progress feature, or is it fine as it is?

romasku:
> Would you like me to indicate that this is a work-in-progress feature, or is it fine as it is?

I think it's fine as it is. Device `MODULE_GIRIER_TS130F_2GANG` is marked as in progress; it should be enough.

LorandBiro:
Thanks! No further changes planned, this is ready to merge.

PR #294, opened 2026-01-17
Try add BY25Q80AW support for ZBMINIL2

romasku:


aschmitz:
My impression is that this PR exists because some ZBMINIL2 devices have different flash memory attached to them? I would not have thought to investigate that discrepancy, but flashing the "released" versions of the bootloader and end device / router firmware didn't work on my ZBMINIL2 (no flashing light, didn't show up in Zigbee), but erasing it, flashing this, and then flashing the end device firmware did! The device now connects via Zigbee and basic functionality works - thanks for this fix! (And for publishing it a few days before I got a proper SWD setup. :wink:)

I'm still having some issues where the quirks file doesn't seem to be being applied to the switch, but even having it working is an improvement over where it was before, and I'll try to debug that separately.

(If it's helpful to identify which devices are affected, I can provide any debugging information you'd like. The ZBMINIL2s I have were ordered in January and April of 2025 from ITEAD US.)

andrei-lazarov:
Yes, there are at least 2 flash chips for this device.

You're the second person to mention ZHA quirks not working _for the SONOFF.._
So probably not an issue on your side. On Z2M it should work.

Also OTA updates aren't confirmed yet on this variant.
Wait a bit before installing the device in the wall 🙂

aschmitz:
So I had apparently forgotten to save the `configuration.yaml` file with the `zha:` section in it before restarting, so now I have the quirks at least showing up.

Trying to set the relay mode, however, gives me "Failed to perform the action select/select_option. Failed to write attribute relay_mode=<RelayMode.Detached: 0>: <Status.UNSUPPORTED_ATTRIBUTE: 134>", so I'm not entirely sure what's going on there. Is there an existing issue for ZHA quirks on the SONOFF device?

(I'm currently using the end device firmware because turning the switch off with the router mode caused the device to restart. I don't really intend to have the relay off most of the time - I'm mostly just using it as a detached smart switch to control a Zigbee light - but I figured it would be a good option to have, and the end device firmware might be less likely to restart without a neutral.)

andrei-lazarov:
The quirks issue was mentioned on Discord. 
We use it for faster troubleshooting.

Yes that's the same error.

- edit: https://github.com/romasku/tuya-zigbee-switch/pull/305

PR #295, opened 2026-01-19
Development container

LorandBiro:
Hey,

I rely heavily on VS Code devcontainers because I use both Linux and Windows machines. I created a devcontainer configuration for this repository that allows VS Code to set up the entire development environment automatically, without requiring any changes to the host system.

This would let contributors simply click "Reopen in Container" in VS Code, and all `make` commands would work out of the box. Is this something you would be interested in? If so, I could update the documentation as well.

For Cursor, I had to use a different extension and install `clangd`, along with a `.clangd` configuration file, because Microsoft does not allow Cursor to install their own extension. Installing the Clangd extension directly would work for both VS Code and Cursor, but the official Microsoft extension appears to be much more popular, so the current setup uses Clangd only for Cursor. I do not have professional experience with C development, so I am not sure what people typically use in projects like this, I hope this make sense.

### DIFF
--- a/docs/contribute/building.md
+++ b/docs/contribute/building.md
@@ -80,7 +80,9 @@ They can easily be adapted for other distributions. (Please share your scripts)
 
 ### 🐳 Dev Container build
 
-As an alternative, dev containers provide a quick local setup without manually installing dependencies on your machine. They work across all platforms that support Linux containers and provide a consistent, isolated development environment. The container automatically installs all required dependencies and tools, configures VS Code extensions (C/C++ IntelliSense, Hex Editor, YAML), and activates the python virtual environment.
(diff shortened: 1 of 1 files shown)


LorandBiro:
On second thought, I removed Clangd for now. It feels overly complex. It requires an APT package, a separate extension with different behavior, and a `.clangd` config file that does not support relative paths. That could easily become a PITA if someone wants to use Clangd outside of the dev container. It works, but it feels quite dirty compared to the plain VS Code solution.

andrei-lazarov:
Yeah this is cool 🙂 You may also add YAML and Hex Editor extensions

romasku:
Thank you!

PR #296, opened 2026-01-20
Add HOMEd extension generator

cetygamer:
This PR adds generator for HOMEd.

HOMEd is a very light weight home automation framework. It can be run on very low-end hardware, such as IoT devices, routers and normal home server PCs. It has a Zigbee service that can be expanded with custom device descriptions similar to zigbee2mqtt. Extensions have JSON format which makes it easy to generate them with Python data structures.

Home page: https://wiki-homed-dev.translate.goog/?_x_tr_sl=ru&_x_tr_tl=en&_x_tr_hl=ru&_x_tr_pto=wapp

The only thing that is not supported (for now) is device string as HOMEd does not support long str ZCL type yet. Everything else, including binding and reporting is implemented.

Screenshots (1-gang module on the left, 2-gang switch on the right):
homed


andrei-lazarov (approved):

romasku:
Looks great, thank you!

PR #305, opened 2026-01-27
🚑️🔨 Fix ZHA quirks for SONOFF

andrei-lazarov:
SONOFF gets UNSUPPORTED_ATTRIBUTE on ZHA.
Supposedly this fixes it


romasku:
Thank you!

Issue #306, opened 2026-01-27
Add Support Module Tuya NovaDigital MS105-ZB


gui9394:
Hello everyone!

Thank you again for the project.

I'm contributing again with the configurations for the NovaDigital MS105-ZB module.
I tested everything and it worked perfectly: reset button, relay, switch, and LED.
I identified the pins using a multimeter.

[Pull Request](https://github.com/romasku/tuya-zigbee-switch/pull/307)

Module Photos:

<img src="https://github.com/user-attachments/assets/17213662-0dce-4aff-a2c8-9aa5a83e19c0" align="top" width="20%"/>
<img src="https://github.com/user-attachments/assets/601f63ec-e777-43b3-8b3e-97d4287ac9fa" align="top" width="20%"/>
<img src="https://github.com/user-attachments/assets/3e5e5e59-3ca0-49b2-b82d-f71abf5be178" align="top" width="20%"/>
<img src="https://github.com/user-attachments/assets/c3e10b91-ca04-4cc8-8978-d076680de667" align="top" width="20%"/>
<p></p>

ZHA Prints:

<img width="919" height="841" alt="Image" src="https://github.com/user-attachments/assets/632bd78f-a25c-4466-8ed1-1fe8d7dfa667" />

<img width="919" height="489" alt="Image" src="https://github.com/user-attachments/assets/0aa15413-40e0-47e5-b93f-d515febe6ae8" />

andrei-lazarov:
Added to main. Thanks!

PR #307, opened 2026-01-27
Add support module Tuya NovaDigital MS105-ZB

gui9394:
[Issue](https://github.com/romasku/tuya-zigbee-switch/issues/306)


andrei-lazarov:
Hi, thanks again! 🙂 

You didn't choose a `firmware_image_type`, so **your device will not receive updates.**

You can change it after installation in Z2M, but not in ZHA.
I recommend you switch to Z2M, you will be able to change the `config_str` without flashing again.
It's also useful for finding pins easier.

I will give it some `firmware_image_type` and run the build, then you can flash it again if you want.

PR #336, opened 2026-02-10
✨ Respond to MoveToLevelWithOnOff

andrei-lazarov:
Device now responds to *MoveToLevelWithOnOff*
  - Level = 0 -> Turn off relay
  - Level > 0 -> Turn on  relay

Tested on Telink (router) board and stub device.
I don't think anything extra is needed for Silabs.

@LorandBiro would this conflict with covers in any way? 😬 
They use entirely different commands, right? 
Does the relay stop reacting to normal on/off/toggle commands if you put it in cover mode?

### DIFF
--- a/docs/changelog_fw.md
+++ b/docs/changelog_fw.md
@@ -8,8 +8,13 @@ Please describe what you are working on:
 
 ### Features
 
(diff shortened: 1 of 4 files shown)


LorandBiro:
@andrei-lazarov Nope, there's no conflict with the cover feature here. Covers use different commands, and they do not respond to any switch commands.

PR #351, opened 2026-02-20
feat: make multi-press factory reset count user-configurable

youtux:
## Summary

Allow users to configure how many rapid button presses trigger a factory reset, or disable the feature entirely. Previously this was fixed at 10 presses with no way to change it.

A new `multi_press_reset_count` setting is exposed in both Zigbee2MQTT and ZHA. Set it to any value between 1-255, or 0 to disable multi-press factory reset completely. The default remains 10, so existing behavior is unchanged. The setting is saved to flash and survives power cycles.

### Implementation

- New writable attribute `multi_press_reset_count` (`0xff02`, `uint8`) on the Basic cluster (endpoint 1)
- New NVM persistence module (`device_params_nv.c/.h`) with key `NV_ITEM_MULTI_PRESS_RESET_COUNT = 33`
- Updated Z2M and ZHA templates to expose the setting

### Design note: separate NVM key

The new value uses its own NVM key rather than being added to the existing `zigbee_basic_cluster_config` struct. Adding a field to that struct would change its `sizeof`, causing `hal_nvm_read` to fail on devices upgrading from older firmware - losing the saved LED state. A separate key ensures a clean upgrade path with no migration needed.

### Question for maintainer

`custom-zcl.xml` / `zcl_config.zap` only define `device_config` (`0xff00`), but `STATUS_LED_STATE` (`0xff01`) and all other custom attributes are registered purely at runtime via `SETUP_ATTR()`. I followed the same runtime-only pattern. Should all custom attributes be declared in the ZCL XML/ZAP files instead?


andrei-lazarov:
This is a long overdue feature, thank you for taking the time to add it! 🙂 

Make sure you:
- **fix formatting with `uncrustify.cfg`**
- **prepare `changelog_fw.md`**
- also update homed extension (or note somewhere that it's todo)
- update readme if needed
- update docs if needed
- test the firmware on a real device

I'm not sure about the NVM situation and the silabs files.

I think the intended way is to use the mechanism in `nvm_migrations.c`
But your solution is also nice. 

Let's wait for romasku's opinion.
Though I am worried about the PRs piling up while he is away.

Also, in the future, I think we should move everything custom to a `manuSpecificRomasku` cluster.

youtux:
Thanks for the review! Here's what I've addressed so far:

- **Uncrustify**: Applied formatting with `uncrustify.cfg` to all changed files
- **Changelog**: Added entry to the WIP section in `changelog_fw.md`
- **HOMEd extension**: Updated `make_homed_extension.py` to expose `multiPressResetCount` for all devices, regenerated `tuya-custom.json`
- **Readme**: Updated the reset options line to reflect configurable count

**Re: NVM approach** — I saw `nvm_migrations.c` but since there are no actual migrations yet and the separate-key approach avoids needing a version bump, I went with that. Happy to switch to the migration approach if romasku prefers.

**Re: `manuSpecificRomasku` cluster** — Makes sense as a future direction. Out of scope for this PR though.

**Testing**: I have an AVATTO ZWSM16-2 and triggered a firmware build on my fork — will OTA and test on the real device once it completes.

Full disclosure: this entire PR was built with Claude Code (Copilot agent mode). I don't have experience with embedded/Zigbee C development, so please do scrutinize the implementation carefully.

youtux:
Tested on a real AVATTO ZWSM16-2 (TS0002, `_TZ3000_mtnpt6ws`) via OTA — works as expected:

- `multi_press_reset_count` shows up in Z2M after flashing (no re-interview needed)
- Value persists across power cycles
- Value persists across factory reset (same as other app NVM settings like LED state — intentional, since `nv_resetToFactoryNew` skips `NV_MODULE_APP`)
- Setting to 0 disables multi-press factory reset

mrmaximas:
@andrei-lazarov How about implementing the pairing mode through multiple power cycles?
For example, I have several relays without switches that are difficult to access under the ceiling.
It would be good for everyone.
Thanks!

romasku:
Thank you for contribution!
Can you please drop "Update firmware files ..." and "Freeze links ..." auto generated commits from this branch, please? We'll do a new build after merging this into main.

youtux:
Yes, I plan to get back to this tonight.

I just noticed though that in home assistant the new field doesn’t display the current setting, it just says unknown:

image

youtux:
also, it shouldn’t be a slider, but just a number you can type

andrei-lazarov:
Look at 'long press duration' (unit16) and 'level move rate' (uint8).
uint8 is displayed as slider. uint16 is with typing.
I guess it's ok like this.

Try to read the value in Z2M (hit the blue arrows next to it). Or just change the value.
It should also update in HA

youtux:
@romasku I deleted those commits

youtux:
@romasku would you be able to check this again?

youtux:
@romasku any chance we can get this merged?

romasku:
Yes, sorry for the delay. I'll look into it.

romasku:
Just checked and all looks nice, thank you very much!

youtux:
@romasku any plan to have this released?

PR #354, opened 2026-02-25
Add support for _TZ3000_q8r0bbvy LVGESS WH-D02

kirill190497:
Add support for _TZ3000_q8r0bbvy LVGESS WH-D02


andrei-lazarov:
Thank you

Issue #373, opened 2026-03-19
Add support NovaDigital ZTS-3W


gui9394:
Hello everyone!

I'm contributing again with the settings for the NovaDigital ZTS-3W wall switch.

I tested everything and it worked perfectly: touch buttons, relays, LEDs, and the reset function by pressing it several times.

I identified the pins using a multimeter.

[Pull Request](https://github.com/romasku/tuya-zigbee-switch/pull/374)

Switch Photos:

![Image](https://github.com/user-attachments/assets/aedc8415-7a04-4b60-a86a-b2c4e98d1d4f)
![Image](https://github.com/user-attachments/assets/284d9dc8-8ee7-4a75-85b2-f3311e5dd925)
![Image](https://github.com/user-attachments/assets/7e63f9c1-efe4-4e18-b60f-92cdee1e8d42)
![Image](https://github.com/user-attachments/assets/0e3f2bf3-5562-47a5-9bad-e8fa7ebe5240)

Zigbee2MQTT:

andrei-lazarov:
Added

PR #374, opened 2026-03-19
Add support Switch NovaDigital ZTS-3W

gui9394:
[Issue](https://github.com/romasku/tuya-zigbee-switch/issues/373)


andrei-lazarov:
Thank you!

PR #375, opened 2026-03-22
Update device_db.yaml - add Moes MS-104CZ end device option (pfc7i3kt)

Hydron-4hv:
Add end_device build option for Moes MS-104CZ (pfc7i3kt) to improve network reliability where it has low signal strength


andrei-lazarov:
Added and started the build

Issue #340, opened 2026-02-10
TS0001 - _TZ3000_kycczpw8 - WHD02


asbachb:
<img width="2299" height="611" alt="Image" src="https://github.com/user-attachments/assets/747cacb4-05c6-4234-9249-2d053ba9817a" />

Zigbee Model: TS0001 (_TZ3000_kycczpw8)
Model: WHD02

Pinout:

| Function | Pin |
| --- | --- |
| Switch | B5 |
| Button | C3 |
| LED | C2 |
| Relais | D2 |

<img src="https://github.com/user-attachments/assets/31e3dfdb-6850-49f9-99df-993282d3824f" align="top" width="32%"/>
<img src="https://github.com/user-attachments/assets/a6c830c6-2b7d-4428-bd28-2a5e077c573f" align="top" width="32%"/>
<img src="https://github.com/user-attachments/assets/2bedd100-8921-4f48-8b04-543d3bb19cfc" align="top" width="32%"/>

andrei-lazarov:
Thank you! This also helps @bordelloweb

Build finished on the main branch. Follow the updating guide

PR #386, opened 2026-04-03
Add support for Mini Smart Switch iHSW02

dexxter00:
This is no name brand smart switch labeled "Mini Smart Switch", model "iHSW02". Outer case is 1to1 compared to Aubess WHD02. The board inside is different layout with the same size, cutouts for the case and connector. It's using ZTU module.
Tested on my device, everything is working fine.

### DIFF
--- a/device_db.yaml
+++ b/device_db.yaml
@@ -202,33 +202,6 @@ BOARD__ZIGBEE_ZTU_2:
   info: Supported
   threads: null
   store: https://www.aliexpress.com/item/1005007599675462.html
-MODULE_MSMARTSW_TS0001:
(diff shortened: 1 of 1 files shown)


andrei-lazarov:
Thanks you!
- so it's similar to https://github.com/romasku/tuya-zigbee-switch/issues/340

dexxter00:
> Thanks you!
> 
>     * so it's similar to [TS0001 - _TZ3000_kycczpw8 - WHD02 #340](https://github.com/romasku/tuya-zigbee-switch/issues/340)

Thank you for merging. Looking at the pictures from #340 , the pcb and case are exactly the same.

Issue #171, opened 2025-10-12
TS0044 4gang (scene remote) switch - custom firmware possible?


ThePegass:
Hello,

I have recently bought TS0044:
https://www.zigbee2mqtt.io/devices/TS0044.html
https://www.zigbee2mqtt.io/devices/TS0044_1.html

<img width="895" height="193" alt="Image" src="https://github.com/user-attachments/assets/87677b4c-5eee-4193-b6a2-32e10a9d8d1a" />

Zigbee module used:
 - **Wall switch (TS0044): ZS3L**
  ![Image](https://github.com/user-attachments/assets/250048c7-240a-4f92-bbfe-206da12fda19)
 - **Remote control (TS0044_1): ZT3L**
 ![Image](https://github.com/user-attachments/assets/9e2314e8-17b6-46d2-b833-832b3106eacf)

All of them works fine, however there is short delay between button press and action response in Z2M I would like to eliminate. The delay is most likely because these devices support also doubleclick and it waits 0.5s if second click occurs (double click action detected). This cannot be normally changed. I was looking for alternative custom firmware and found "[tuya-zigbee-switch](https://github.com/romasku/tuya-zigbee-switch)"

Based on readme, module ZT3L is supported, but this TS0044 device is not in the list of supported devices. Would it be possible to create custom firmware for TS0044_1? 

Appreciate any help or guide.
Thanks

andrei-lazarov:
Hi, thanks for your interest in the project! 
Currently the firmware only supports Telink-based chips (address starting with 0xa4c138) like the the ZT3L. 

However Romasku is working on Silicon Labs support (like the ZS3L). Some info is in https://github.com/romasku/tuya-zigbee-switch/issues/14#issuecomment-3357588800. 
I also have the 4-gang scene switch so I will help porting and testing. Will let you know when it will be available (not very soon I imagine).

To port the firmware for the compatible remote, we need to know the pinout of the module: which pins of the ZT3L connect to which peripherals (buttons, LEDs, relays).  
It's best to check with a multimeter, but the traces are visible so we can work it out. I see 5 traces. Snap a picture of the other side of the PCB. Some examples here: https://github.com/romasku/tuya-zigbee-switch/issues/145#issuecomment-3303035527, https://github.com/romasku/tuya-zigbee-switch/issues/146#issuecomment-3302750944

[...]

PR #395, opened 2026-04-13
Fix config for REMOTE_IHSENO_TS0044

Haydend:
I've just onboarded some 4 button remotes (https://www.zigbee2mqtt.io/devices/TZ3000_mh9px7cq.html#ihseno-tz3000-mh9px7cq) and noticed the config is wrong

**Summary Of Changes**
- Remove relays - device has none
- Fix Switch 4 Pin - looks like the OG issue (https://github.com/romasku/tuya-zigbee-switch/issues/171
) miss traced the track, it looks like it goes to D7, but it actually runs under the module to B4
- Add BT parameter, to enable battery optimization - This appears to require a pin? I've assigned it the only ADC pin on the board C4, but nothing is connected, so I don't believe it is possible to get the battery level on this device.  (https://developer.tuya.com/en/docs/iot/zt3l-module-datasheet?id=Ka438n1j8nuvu#title-5-Pin%20definition)

---
P.S. This is my first contribution, so I want to say massive thanks for creating this custom firmware, it is excellent! and unlocks many options for a more decentralized smart home.

### DIFF
--- a/device_db.yaml
+++ b/device_db.yaml
@@ -2854,8 +2854,10 @@ REMOTE_IHSENO_TS0044:
   firmware_image_type: 43563
   build: yes
   status: in_progress
-  info: Battery Pin is not connected, so will alway report 100%
-  threads: https://github.com/romasku/tuya-zigbee-switch/issues/171
+  info: Battery Pin is not connected, so will always report 100%
+  threads: 
+   - https://github.com/romasku/tuya-zigbee-switch/issues/171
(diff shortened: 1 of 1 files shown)


andrei-lazarov:
Thank you! Try to put the battery pin on one of the buttons:
`mh9px7cq;TS0044-CUS;LC0;SD2u;SC2u;SC3u;SB4u;BTD2;M;`

PR #396, opened 2026-04-14
Added support for OXT 3CH with N

9mrcookie9:
Ahoj! 
This time i'm pleased to add support for OXT whitelabeled module - 3 channels relays.


andrei-lazarov:
Thank you!

PR #397, opened 2026-04-15
Add BSEED PM outlet 🅲

dbloms:
I got an another model of the TS011F_plug_1_2 power outlet, which will be supported by this new config.


andrei-lazarov:
Thank you again!

PR #406, opened 2026-04-18
add outlet OUTLET_BORUIDAPLS_PM_TS011F

dbloms:
I've added a new devices.  
It has a blue indicator led, which is hardwired, so we can't configure it.


andrei-lazarov:
Reverted to TS011F_plug_1_2 to match ZHC. Looks like they are reusing the firmware from bseed 😢 can't have nice things

Thank you!

PR #408, opened 2026-04-21
Cover position control

LorandBiro:
This PR is the 3rd phase of the window covering feature started in issue #270. It contains position control and tracking.

The branch needs a little more work so it's created as a draft, but builds are available for testing:
- [OTA index](https://raw.githubusercontent.com/LorandBiro/tuya-zigbee-switch/refs/heads/cover-position-build/zigbee2mqtt/ota/index_router.json)
- [Z2M converter](https://raw.githubusercontent.com/LorandBiro/tuya-zigbee-switch/refs/heads/cover-position-build/zigbee2mqtt/converters/switch_custom.js)


LorandBiro:
I added an "overrun" feature, which lets you press open or close again at the end position to run the motor for another 3 seconds. I did this because calibration was frustrating when the controller thought it had reached the end position, but the cover hadn’t actually reached it physically. The only way to fix that was to send the cover all the way in the opposite direction and then back again. This helped a lot while I was calibrating the dead zones.

I’m also considering adding a fixed 1-second overrun whenever we’re at an end position, to improve accuracy when the cover only moves between 0% and 50% without ever reaching 100%. A fixed overrun would ensure that 0% truly hits the physical end position.

Any feedback/idea on how to improve this is welcome.

LorandBiro:
Ok, so the question about the "overrun" feature seems to be clarified. I added a 3-second overrun each time the cover is sent to an end position. This allows for realignment in cases where the cover is only used at one end of its range. If both the open and close times are set to 0, manual mode is activated. In this mode, there is no position tracking or control. Based on testing so far, this provides an easy way to set up and calibrate the covers.

LorandBiro:
I’ve finished the PR, but there are two points where I need help:

- I updated `MODULE_TUYA_NOVATO_TS130F` to enable OTA updates. @DawidLbd, @andrei-lazarov - could you test and verify that OTA is working and that my changes are correct?
- I had to implement a workaround to retrieve the argument from the `GoToLiftPercentage` command. It doesn’t look great. Does anyone have an idea why `cmd_incoming_from_addr_info` is not working here? See [zigbee_zcl.c](https://github.com/romasku/tuya-zigbee-switch/pull/408/changes#diff-08e1d8d16a0b78a2eced816d995d6917e7b1c9dfa4d9b05cb446bd3baf479a30R109)

DawidLBD:
I performed an OTA test with the custom FW, and it loaded without any issues. The config string is correct.

Haydend:
I did some investigation into why cmd_incoming_from_addr_info isn't working for you.

I believe it's because of zcl_window_covering_custom.c. This memory stuff is a bit beyond me at the moment, so I don't really understand what's happening there. But I used zcl_level.c from the Telink library as an example and reimplemented it. That appears to have fixed it.

I have a branch with the changes you can take a look at. I've run it on a device and it worked.
The key files:
https://github.com/Haydend/tuya-zigbee-switch/blob/dunnicliffeh/cover-position-investigation/src/telink/custom_zcl/zcl_window_covering_custom.c
https://github.com/Haydend/tuya-zigbee-switch/blob/dunnicliffeh/cover-position-investigation/src/telink/hal/zigbee_zcl.c

It feels like Telink are doing something weird/wrong here, but it might be best to follow their lead to avoid confusion about why this one callback works differently.
Hopefully this helps someone who's better at C to understand what's happening here.

Also, have you done all this using only OTA? That's hardcore! I would of bricked so many devices.

Haydend:
TL;DR In zcl_window_covering_custom.c pass a pointer to zclIncoming_t->addrInfo, instead of building your own zclIncomingAddrInfo_t. 
`return(incoming->clusterAppCb(&(incoming->addrInfo), incoming->hdr.cmd, payload));`

Slept on this, and I think I understand it a bit more now. The Telink cmd handlers all pass a pointer to the addrInfo (zclIncomingAddrInfo_t ) member of the zclIncoming_t struct they receive. cmd_incoming_from_addr_info is some C magic which gets you a pointer to the whole zclIncoming_t struct, from a pointer to a member of that struct. In your custom cmd handler, you were building a fresh zclIncomingAddrInfo_t struct. The pointer to this fresh struct is in no way connected to the incoming zclIncoming_t, thus cmd_incoming_from_addr_info doesn't work. 

Still not 100% clear on how cmd_incoming_from_addr_info works. It appears to also be known as the container_of macro and is used a lot in the linux kernel. Something to do with a struct being contiguous memory, therefore if you have a pointer to something in that struct you can travel along the memory to find the pointer to the whole thing. Very clever.

See https://github.com/Haydend/tuya-zigbee-switch/blob/dunnicliffeh/cover-position-investigation/src/telink/custom_zcl/zcl_window_covering_custom.c

LorandBiro:
@Haydend Thank you so much for looking into this, it felt impossible to figure out without serial access or access to the Telink source code. I tested the changes locally, and everything seems to work correctly.

> Also, have you done all this using only OTA? That's hardcore! I would have bricked so many devices.

Yeah, I did everything over OTA, the first update was nerve-wracking :D But I got lucky and didn’t brick any devices.

FrancescoTalotta:
Hello,

Thank you very much for your help.

I have six of these curtain modules purchased from AliExpress: https://alishort.com/IebOn

I was wondering whether it would be possible to include the corresponding firmware support directly in the main build.

I noticed that there are two links available: one for the OTA firmware and one for the Zigbee2MQTT converter. Could you please confirm whether these are already updated to the latest version?

Thank you in advance for your support.

Francesco

LorandBiro:
@FrancescoTalotta 

> Hello,
> 
> Thank you very much for your help.
> 
> I have six of these curtain modules purchased from AliExpress: https://alishort.com/IebOn
> 
> I was wondering whether it would be possible to include the corresponding firmware support directly in the main build.
> 
> I noticed that there are two links available: one for the OTA firmware and one for the Zigbee2MQTT converter. Could you please confirm whether these are already updated to the latest version?
> 
> Thank you in advance for your support.
> 
> Francesco

Hey,

I also use those curtain modules, it should be working perfectly.

Some functionality is already merged to main. If you can live without the position control, you can already update your devices by following the instructions in the documentation. This PR will be merged soon, then an official firmware upgrade will add position control.

The converter and OTA files are not updated in this branch. If you want to test the changes before the merge, you can get the binaries and the converter from my build branch: [cover-position-build](https://github.com/LorandBiro/tuya-zigbee-switch/tree/cover-position-build)

You can get the binaries directly from there, or you can follow the documentation and change the URLs. Please note that you'll have to wait for this build before the files are updated on the build branch: https://github.com/LorandBiro/tuya-zigbee-switch/actions/runs/25492620252

FrancescoTalotta:
Hello,
Thank you very much for the binaries, and really nice work indeed!
I flashed the following firmware: Firmware ID: 1.1.2-0564bee3 20260507 Firmware version: 285356037

I have not tested all the functions yet, but the switch already works much better with this new firmware.
However, I noticed two issues. It seems that I cannot invert the motor direction, meaning the UP/DOWN functions cannot be swapped. In order to see the cover position percentage, I need to manually click the small refresh button on the right to update the position. With the stock firmware, the position was updated automatically by sending MQTT messages. This was very useful, especially to monitor remotely what was happening with the cover when I was away from home.

Thank you again for your work and support.

EDIT: Probably it was me or something wrong with Z2M. This morning I opened Z2M again and the motor direction and cover position percentage were working very well. The position is updated automatically. 

EDIT2: Ok found the issue. I forgot to Reconfigure the devices after firmware upgrade. Now it works very well.

Francesco

Todelec:
Hi,
great work you're doing here.

I first updated my 2-gang switches and then came across your roller shutter branch, which I flashed immediately.
So far it’s working excellently!

I’m already looking forward to the position display in the Home Assistant frontend.

Are you planning to implement direct binding?

LorandBiro:
@FrancescoTalotta Thank you very much for testing the branch and sharing your solution.

@Todelec Happy to hear that! Could you clarify which firmware version you deployed to your shutter devices? Home Assistant should already display the position properly if you're using the build from my branch. If the position is available in the Z2M UI but not in Home Assistant, please share the details so I can investigate further.

Direct binding is already available in the official build. You can set up the switches of a cover module to control other cover modules. Is this what you meant by direct binding?

Todelec:
Hi,

Thank you for your quick reply!

I’m using version 1.1.2-0564bee3. At first, there were no entries in the Z2M reporting tab, but after adding them, the position is now displayed correctly in the frontend. That was my mistake.

Regarding binding:
I would like to bind a TS0002-GIR-zmy4lslw device to a cover module. However, when selecting the cover module as the target, no clusters are available to choose from. I only see endpoints 1 and 2, plus cover_switch and cover.

I was able to successfully bind two TS0002-GIR-zmy4lslw devices together. 
However, I’m still new to this topic and learning. Any advice would be appreciated.

Thanks in advance!

LorandBiro:
@Todelec if there were no entries in the reporting tab, you should probably re-interview and reconfigure the device, that usually helps. Also, binding has a really bad user interface in Z2M. I just tried binding between two Girier modules, and when I chose the second endpoint, I could choose the cluster as well. After that, binding worked properly.

image

yarfalksol:
@LorandBiro thank you very much for your effort, great work!
I'm currently using a few Wi-Fi shutter modules, but I'm planning to switch to Zigbee, as Zigbee is my primary choice for switches and dimmers
I've tested your latest build `0564bee3` on `GIRIER_TS130F_1GANG/dwytrmda` with great success. I have two comments:
- my coverOpenTime is 66 seconds and coverCloseTime is 62 seconds, but it seems the maximum for both in the current build is 60 seconds. 60 seconds works fine, but 66 only lasts a few seconds. I found your statement: _Add coverOpenTime and coverCloseTime converter helpers (numeric, 0–**60**000 ms, config category)_ - could the upper limit be raised?
- my Wi-Fi shutter modules support slat angle positioning. Do you plan to introduce this feature as well?

Todelec:
> @Todelec if there were no entries in the reporting tab, you should probably re-interview and reconfigure the device, that usually helps. Also, binding has a really bad user interface in Z2M. I just tried binding between two Girier modules, and when I chose the second endpoint, I could choose the cluster as well. After that, binding worked properly.


@LorandBiro 

Thank you very much for your support.

What I was trying to do apparently cannot work.
I wanted to bind a regular switch with the cover module.

In the meantime, I found out that this cannot work because the switch does not have a closuresWindowCovering client cluster.

If I use a cover module instead, I can select the closuresWindowCovering cluster as well, just like in your screenshot.

So that’s how I’m going to rebuild it now — of course with your firmware ;)

BR and thanks a lot!
Tod

LorandBiro:
@yarfalksol Sorry I missed your message last week. The open and close time is currently limited to 300 seconds:
[https://github.com/romasku/tuya-zigbee-switch/pull/408/changes#diff-a5fccb8a1b662749470d1d475d4735323a99640a3796826a68375da1a1c8d0baR372](https://github.com/romasku/tuya-zigbee-switch/pull/408/changes#diff-a5fccb8a1b662749470d1d475d4735323a99640a3796826a68375da1a1c8d0baR372)

Where did you see that the open and close time is limited to 60 seconds? I looked through the git history and the code, but I couldn't find it.

I don't plan to add support for slat angle. This "small" contribution already turned out to be much bigger than I anticipated :)
But I do recommend creating an issue, since others have already mentioned this requirement.

yarfalksol:
@LorandBiro seems cover_travel_time() returns uint16_t, causing overflow for travel times > 65 s                  
                                                         
cover_travel_time() in src/zigbee/cover_cluster.c (line 54) stores the result of open_time * 100 (or close_time * 100) in a uint16_t return value. Since open_time/close_time are stored in tenths of a second (zigbee2mqtt scale: 10), a 66 s travel time is stored as 660, and 660 × 100 = 66 000 silently wraps around to 464 (66000 mod 65536). This makes the motor stop after ~3 s instead of running the full travel time.
                    
The overflow threshold is 65535 / 100 = 655 tenths = 65.5 s. Values at or below 65 s work correctly; 66 s and above are broken. The same issue affects both open_time and close_time.
                                                                                                       
Seams cover_travel_time() on line 54 have to be changed from uint16_t to uint32_t. All call sites (lines 113, 152, 284) already assign the result to uint32_t, so no other changes are needed.

Once again thank you very much for your effort.

LorandBiro:
@yarfalksol Oh, I understand now. Thank you so much for investigating this issue in so much detail. I applied the fix and I marked you as co-author. The new build should be available soon.

mikka005:
Hello.
Is this compatible with MOES ZRS-SR-EUC curtain/shutter switches? I really need a solution to opeate one shutter motor with two or more switches and make them in same state (i mean % position) Cannot do it with automation and can not bind switches (its unavaible). Any opinion?

LorandBiro:
@mikloskatay Is this the device you have? https://moeshouse.com/products/zigbee-star-ring-curtain-switch
It looks like the device db doesn't have this yet, so you would have to create a new issue to request/add support to that device.

The firmware on this branch supports your use-case, it's possible to bind multiple curtain/shutter switches together and control the same motor. It doesn't really control the percentage, it just sends open/close/stop commands over the network.

mikka005:
> @mikloskatay Is this the device you have? https://moeshouse.com/products/zigbee-star-ring-curtain-switch It looks like the device db doesn't have this yet, so you would have to create a new issue to request/add support to that device.
> 
> The firmware on this branch supports your use-case, it's possible to bind multiple curtain/shutter switches together and control the same motor. It doesn't really control the percentage, it just sends open/close/stop commands over the network.

Hello Lorand,
Yes it is! I have a lot of them. What shall I do exatly?

Thanks,
Miklos

ChriWinter:
> Hello. Is this compatible with MOES ZRS-SR-EUC curtain/shutter switches? I really need a solution to opeate one shutter motor with two or more switches and make them in same state (i mean % position) Cannot do it with automation and can not bind switches (its unavaible). Any opinion?

> @mikloskatay Is this the device you have? https://moeshouse.com/products/zigbee-star-ring-curtain-switch It looks like the device db doesn't have this yet, so you would have to create a new issue to request/add support to that device.
> 
> The firmware on this branch supports your use-case, it's possible to bind multiple curtain/shutter switches together and control the same motor. It doesn't really control the percentage, it just sends open/close/stop commands over the network.

Hi @LorandBiro & @mikloskatay 

I also have many of these Moes curtain switches (ZS-SR-EUC; _TZE204_srmahpwl). As I’ve installed a bunch of other (supported) Moes devices and as i'm super happy about this custom firmware, I took a look at the board and the pinout for the ZS-SR-EUC. Unfortunately, it seems the built-in ZT2S only acts as a wireless interface for an MCU (CS95F), which does the actual work – not a single LED, button or relay is connected directly to the chip. I have therefore not submitted a request for a new device. However, if anyone thinks the custom firmware could still be applied here, I’d be happy to submit a new request and provide images and the pinout.

andrei-lazarov:
No word from romasku, but this is long overdue, so I'm merging. 🙂
Many thanks to everyone involved in developing and testing!

PR #417, opened 2026-05-06
Add latching pulse duration config

Aranot:


andrei-lazarov:
Thank you! So is this enough to fix your switch?

Aranot:
Yes, so far there haven't been any problems with power supply even if I spam switching

andrei-lazarov:
Thanks!

PR #423, opened 2026-05-20
Config for BSEED Echo 3gang L-only

dexxter00:
Hi, recently purchased this switch and I was able to make it work. I was not able to find the exact switch on the BSEED website(no-neutral version), so I'm reusing  names from similar switches.
Everything is working fine.


andrei-lazarov:
Thank you! I also fixed Z2M detection of the original firmware:
- https://github.com/Koenkk/zigbee-herdsman-converters/pull/12259

PR #428, opened 2026-05-28
Fix Tuya DIN circuit breaker

bb:
DIN_RELAY_TUYA_TS011F_PM
`_TZ3210_vbfp8eyv`
fixed after confirming on physical device

cover
front
back
bottom


andrei-lazarov:
Thanks! I'd like to keep it as `in_progress` while it's missing power measurements

PR #437, opened 2026-06-06
Fix Silabs HAL: also send commands to group bindings

eakorolev:
Discovered on **Moes ZT-B-EU2** (TS0042) that buttons bound to a Zigbee group emit nothing on press, while unicast bindings work fine.

Turns out `hal_zigbee_send_cmd_to_bindings()` only calls `sl_zigbee_af_send_command_`**`unicast`**`_to_bindings()` — group entries have `binding.type == SL_ZIGBEE_MULTICAST_BINDING` and get filtered out by the unicast walk. The SiLabs framework itself handles this by calling both `_unicast_to_bindings()` and `_multicast_to_bindings()` back-to-back after a single `fill_external_buffer` (see [`reporting.c:367-381`](https://github.com/SiliconLabs/simplicity_sdk/blob/v2025.6.2/protocol/zigbee/app/framework/plugin/reporting/reporting.c#L367-L381)) — mirrored here.

603920460-a8c8f742-179f-47c5-a8e4-45773f1242bc

603921718-dd712d67-f677-4730-a74f-12ec850142b9
603921850-42794ad0-4352-47fe-8d51-771a7de4f961


eakorolev:
@romasku @andrei-lazarov should I update firmware version right now, or CI will do it after merge?

andrei-lazarov:
This looks like a big omission, how hasn't anyone reported yet? 😯 Thanks for fixing

Not sure what to do while romasku is inactive.. I don't want to update source code and release without him

Also I've removed the image metadata (location) for your privacy...
(GitHub uploads in original -> you should do some compression before-hand)

eakorolev:
I'm just playing around with a few devices and trying to make my smart home work exactly the way I want. Native Zigbee bindings are always better than integrations through HA, so I'm using them everywhere it's possible =)

A big update with support for a separate long-press endpoint is almost here, but one of my (battery) devices suddenly dropped off the network, so I'm doing some field testing before publishing the PR.

Thanks for the metadata, I wasn't worried about it.

ohmstance:
Hey I built your PR and binding to group works now. Thank you!

romasku:
Great catch, thank you!

PR #450, opened 2026-06-20
Add inverted flag to network indicator LED

dexxter00:
Hi, I noticed that when network indicator switch is off, the LED is actually lit. It should be inverted.


andrei-lazarov:
Thank you!

Issue #422, opened 2026-05-19
Sonoff ZBMINIL2 stays completely dead


Remboooo:
I've followed the guide at [docs/flashing/zbminil2.md](https://github.com/romasku/tuya-zigbee-switch/blob/main/docs/flashing/zbminil2.md), but ran into a few problems.

TL;DR: it seems to flash the bootloader and end-device/router firmware just fine, but after that the device seems completely dead. It does not blink the LED, respond to button presses or join the network.

Firstly, the Commander GUI does not work at all for me. It always comes back with `WARNING: Could not connect to target device`. I'm confident I selected the right JTAG adapter, I set it to SWD and tried different speeds, but it just keeps blinking the J-Link LED for a while and coming up empty.

Next I tried the Commander CLI which seems much more successful.

My first attempt was:
```
.\commander-cli.exe device recover --device efr32mg22
Recovering "bricked" device...
Resetting device...
Attempting to connect to DCI...
Success after 1 attempts (2 ms)
Running device erase command...
Success after 1 attempts (85 ms)
Resetting device...
Device recovery successful.
DONE
```
Excellent. Next I tried flashing the bootloader and end device firmware:
```
.\commander-cli.exe flash C:\users\rembr\Downloads\efr32mg22_bootloader_module_sonoff_zbminil2.s37 --device efr32mg22
Reconfiguring debug connection with detected device part number: EFR32MG22C224F512GN32
Parsing file C:\users\rembr\Downloads\efr32mg22_bootloader_module_sonoff_zbminil2.s37...
Writing 24576 bytes starting at address 0x00000000
Comparing range 0x00000000 - 0x00005FFF (24 KB)
Erasing range 0x00000000 - 0x00005FFF (3 sectors, 24 KB)
Programming range 0x00000000 - 0x00000FFF (4 KB)
Programming range 0x00001000 - 0x00001FFF (4 KB)
Programming range 0x00002000 - 0x00002FFF (4 KB)
Programming range 0x00003000 - 0x00003FFF (4 KB)
Programming range 0x00004000 - 0x00004FFF (4 KB)
Verifying range 0x00000000 - 0x00005FFF (24 KB)
Flashing completed successfully!
DONE
```
```
.\commander-cli.exe flash C:\users\rembr\Downloads\tlc_switch-1.1.2-699fc909.s37 --device efr32mg22
Reconfiguring debug connection with detected device part number: EFR32MG22C224F512GN32
[...]

PR #457, opened 2026-07-10
Bootloader: generic JEDEC SPI-flash fallback for unknown chips

barbalexs:
## Problem

The Silabs (Gecko) spiflash bootloader identifies the external SPI-NOR flash by its JEDEC id (`0x9F`) against a hardcoded table in `getDeviceType()`. If a chip's id is not in the table, `getDeviceType()` returns `UNKNOWN_DEVICE`, `storage_init()` returns `BOOTLOADER_ERROR_INIT_STORAGE`, and the application never boots — the module is **dead after flashing** (no LED, doesn't join).

This recurs per hardware variant. The ZBMINIL2 alone ships with at least three different flash chips, each currently needing its own bootloader build:

| Chip | JEDEC (`0x9F`) | Status on `main` |
|---|---|---|
| Tsingteng TS25Q80UA | `EB 60 14` | in table |
| Boya BY25Q80AW | `68 10 14` | added in #294 (still open) |
| Giantec GT25Q80A | `C4 60 14` | not handled → dead after flash |

This is the class behind #422 and #294.

## Change

Add a **generic JEDEC fallback**: when the JEDEC id is not in the table but is plausible (manufacturer byte not `0x00`/`0xFF`, sane density code), serve it with a generic standard-SPI-NOR profile whose total size is derived from the JEDEC density code (`1 << code`). The command set (`0x03`/`0x02`/`0x20`/`0xD8`/`0x06`/`0x05`), 256-byte page and 4 KB sector are already universal across every chip in the driver, so only the id gate needed generalizing.

Properties:

- **Additive / no regression** — the explicit table still matches first; every already-supported chip is byte-for-byte unchanged. The generic branch only catches table misses.
- **Gated** by `BTL_STORAGE_SPIFLASH_GENERIC_JEDEC` (default **on**; set `0` to require an explicit table match).
- **Safe on absent flash** — a missing/floating chip (bus reads all `0x00`/`0xFF`, or an out-of-range density code) is rejected, so "no chip present" is never mistaken for a valid part.
- The new `GENERIC_JEDEC_DEVICE` enum value is appended **last**, keeping it outside the `ATMEL_*`/`NUMONYX_*`/`ISSI_*` value ranges that `storage_init()` / `storage_eraseRaw()` test for part-specific handling.

This generalizes the per-chip approach and should cover the #422 class, making one-off additions like #294 unnecessary for standard `25Qxx` parts.

## Testing

Built and flashed on a real Sonoff ZBMINIL2 whose flash is a **Giantec GT25Q80A (JEDEC `C4 60 14`)** — deliberately **not** added to the table, so it exercises the fallback exactly as a genuinely-unknown chip would:

- **Before:** dead after flashing — no LED, does not join (the reported symptom).
- **After (this bootloader):** module boots, LED active, and **rejoins the Zigbee network** (confirmed in Zigbee2MQTT).

`bin/efr32mg22_bootloader_module_sonoff_zbminil2.s37` is rebuilt to include the fallback (built with Simplicity SDK 2025.6.2 — feel free to rebuild in your environment).

Refs #422, #294


romasku:
Thanks, this is a great contribution! And sorry for the late reply.

NickHu:
This is fantastic. Does it fix the lack of OTA which was present in the previous BY25Q80AW build?

DJTerentjev:
@barbalexs could you please let me know if OTA works for ZBMINIL2. Because mine is failing.

PR #463, opened 2026-07-16
Add Interlock Mode

gui9394:
# Description

This PR introduces an **Interlock Mode** (mutually exclusive switch mode) for multi-channel switches (devices with 2 or more relay channels). 

When active, it ensures only one relay channel is active at a time. If another channel is commanded to turn on, the previously active channel is turned off first, followed by a configurable delay (in milliseconds) before the requested channel is activated.

### 1. Firmware Logic (`src/zigbee/relay_cluster.c`, `relay_cluster.h`)
* Implemented basic interlock checks inside `relay_cluster_on`.
* Added asynchronous task scheduling via `hal_tasks_schedule` to implement the delay without blocking the CPU.
* Added a `pending_on` state tracking variable to handle race conditions under rapid sequential toggling, canceling pending relay activations when a different channel is selected.
* Forced immediate `OFF` reports (`hal_zigbee_send_report_attr`) when a pending channel activation is canceled, keeping the controller's UI in sync with the hardware.
* Deferred ZCL `report_handler` execution outside of ZCL message processing callbacks to prevent report packet drops inside command handlers.

### 2. Zigbee Custom Attributes (`src/zigbee/consts.h`, `basic_cluster.c`, `basic_cluster.h`)
* Added `0xFF03` (`ZCL_ATTR_BASIC_INTERLOCK_MODE`) - Boolean.
* Added `0xFF04` (`ZCL_ATTR_BASIC_INTERLOCK_DELAY`) - Uint16 (in milliseconds).
* Enabled NVM persistence.

### 3. Integrations & Generators
* **Zigbee2MQTT:** Updated converter templates (`switch_custom.js.jinja`) to automatically expose `interlockMode` and `interlockDelay` on devices with 2+ relays.
* **ZHA:** Added ZHA quirk mapping support in `zha_quirk.py.jinja`.
* **HOMEd:** Extended `make_homed_extension.py`.

### DIFF
--- a/helper_scripts/make_homed_extension.py
+++ b/helper_scripts/make_homed_extension.py
@@ -101,6 +101,13 @@ if __name__ == "__main__":
             "multiPressResetCount": {"type": "number", "min": 0, "max": 255}
         }
 
+        if relay_cnt > 1:
+            basic_custom_attrs["interlockMode"] = {"type": "bool", "clusterId": 0x0000, "attributeId": 0xff03, "dataType": 0x10, "action": True}
+            basic_custom_attrs["interlockDelay"] = {"type": "value", "clusterId": 0x0000, "attributeId": 0xff04, "dataType": 0x21, "action": True}
+            basic_exposes.extend(["interlockMode", "interlockDelay"])
+            basic_options["interlockMode"] = {"type": "toggle"}
+            basic_options["interlockDelay"] = {"type": "number", "min": 0, "max": 5000}
+
         if has_dedicated_net_led:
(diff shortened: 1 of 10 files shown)


falkenhawk:
@gui9394 thank you. I am soo looking forward to see this included in the main build. @andrei-lazarov 🙏 I need it in order not to kill my kichen hood turbine, which never must be supplied with live over more that 1 speed at a time.

May I test the build somehow on a 4ch Avatto ZWSM16-4-ZIGBEE? I guess the definitions need to be adjusted to cover that module?

PR #475, opened 2026-07-25
Add RETLUX RSH30x switches

substanc3-dev:
Adds some switches commonly available in Slovakia and the Czech republic.
_TZ3000_ns91oqaa = 1-gang variant
_TZ3000_uprzndgg = 2-gang variant

On the spritual journey to figure this out I ended up setting up a port of usb2swire to the Raspberry Pi Pico which can be found here: https://github.com/substanc3-dev/pico-usb2swire
Not sure how useful to people this will be but I think it's a pretty nice alternative, especially considering I struggled getting the UART ways to work well.

### DIFF
--- a/device_db.yaml
+++ b/device_db.yaml
@@ -5838,7 +5838,7 @@ SWITCH_RETLUX_RSH301:
   build: yes
   status: fully_supported
   info: Supported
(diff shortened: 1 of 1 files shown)


andrei-lazarov:
Wow, awesome project, must check it out! Thank you!

PR #478, opened 2026-07-27
Fix uncrustify formatting in spiflash storage driver (lint is red on main)

barbalexs:
The `lint` job has been failing on `main` since #457 was merged — see the `test` workflow runs for 13bc4d1 and every push since; 0c8bc30 and 44f74ed before it were green.

Two lines introduced by that PR (mine, sorry) don't match `uncrustify.cfg`:

- `btl_storage_spiflash_configs.h`: alignment of `#define TIMING_ERASE_GENERIC_MAX_MS`
- `btl_storage_spiflash.c`: missing blank line before `#endif`

Running `uncrustify` over `src/` locally, these were the only two failures out of 122 files, so this makes `lint` green again.

Note on reproducing: the CI image (ubuntu-24.04) ships **uncrustify 0.78.1**, and older versions disagree with it — 0.72 (Debian 12) reports failures on files that CI accepts, which is how I missed this in the first place. Worth pinning the version in the workflow if you want the check to be reproducible locally.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


andrei-lazarov:
Thanks!

PR #495, opened 2026-09-01
Propagate relay changes to Zigbee bindings

Gabrielguadag:
Found while binding two relay endpoints together on a secondary-MCU board (built on #484, thanks @charliesjc) — this is base ZCL/binding code with no dependency on that work. A relay's own bindings only ever heard about a state change from a physical key press; a remote command, automation, or UI toggle never propagated. Implemented the missing report function using the same mechanism the switch cluster already uses, with a guard against two bound relays echoing forever. Also registers the relay endpoint as a bind client, not just a server — without it, Z2M refuses to let a relay be a bind source at all. Devices need a fresh interview for that part to take effect.


andrei-lazarov (commented):
When you toggle a relay, it should only send _Report Attributes (OnOff)_ to the bound target.
Not _On / Off_. Why do you want that? It is intentionally only on button press (and different endpoint).

Gabrielguadag:
The main reason is that I want to use them in a 2-way/3-way switch setup. If I toggle Relay A from the coordinator/app, Relay B doesn't change its state because it's only receiving the attribute report instead of an actionable command. I just want to ensure both relays stay completely in sync regardless of how they are triggered (UI or physical press).

Gabrielguadag:
Actually, please ignore my previous comment. I took some time to read more about the ZCL standards and realized my conceptual error. I was trying to use bindings to keep the relays in sync via the UI, when I really should just be using Zigbee Groups. I understand now why this breaks the standard behavior.

andrei-lazarov:
Well it's not a bad idea..
It actually doesn't break anything, since the button and relay are on different endpoints.

But it shouldn't be necessary.
Yes, if controlling from coordinator/remote, you can send the command to both relay A and B, or to a group containing them.

PR #496, opened 2026-09-01
Add three device_db entries for undocumented stock fingerprints

Gabrielguadag:
Found while converting three more units from stock firmware. None of the three fingerprints appear anywhere in this repo or in zigbee-herdsman-converters' own device database — checked directly against both, genuinely new, not just missing locally.

MODULE_TUYA_COMMON2_TS0004 — a 4-gang module with a pin layout not seen on any other TS0004 entry here (relay on A0). Confirmed live against the running device: it reports manufacturer "liygxtcq" exactly as declared, reused deliberately from MODULE_TUYA_COMMON_TS0004 for the same converter shape despite the different wiring underneath.

SWITCH_TUYA_TOUCH_TS0003 — pinout matches SWITCH_AVATTO_TOUCH_TS0003's alt_config_str exactly. Brand left generic rather than assumed, since the matching pinout doesn't establish who sold this specific unit. The real find is the stock_image_type: this unit ignored AVATTO's own OTA offer (43542) and only accepted 43549, the value SWITCH_BSEED_TOUCH_MELODY_TS0003 declares — converted with that image, pinout confirmed only afterward by probing.

SWITCH_TUYA_B_TS0002 — a different physical product from MODULE_EKAZA_TS0002 (a complete 2-gang touch switch panel, not that entry's in-wall retrofit module) that happens to accept the same OTA image. Own pinout, confirmed working.

All three marked in_progress (single unit tested each), no firmware_image_type collisions with any existing entry.


andrei-lazarov:
Thanks! Is there an on-board button on `_TZ3000_2wp3tnwa`?

PR #511, opened 2026-09-16
Fixed config string example in porting.md

labuwx:
The example for `LP<N>` was probably copied over from `D<N>` but was not updated to reference `LP<N>`.


andrei-lazarov:
Good catch, thanks!

PR #514, opened 2026-09-17
relay: re-affirm power-on-behavior shortly after boot

terragady:
## Problem
Power-on behavior "ON" is not always applied reliably after a real power loss on
some devices (Telink), even though the configured startup_mode is confirmed
persisted correctly in NVM. Reported symptom: relay stays off after power-cycle
significantly more often when it was off before the cut, ~50/50 when it was on
before - consistent with a boot-time brownout/reset racing with the relay's
initial drive rather than a logic bug in the startup_mode decision itself.

## Fix
`relay_cluster_handle_startup_mode()` keeps its existing immediate behavior
unchanged, but now also schedules a one-shot, non-blocking recheck
(`STARTUP_RECHECK_DELAY_MS`, 400ms) that re-asserts the same decided target.
This re-assertion is unconditional (not gated on the relay's own on/off
bookkeeping), since a brownout can leave that bookkeeping "correct" while the
relay itself failed to physically engage on the first attempt.

Any real relay state change (command/automation) cancels the pending recheck,
so it can never override a legitimate action taken shortly after boot.

## Testing
Added two tests in `test_relay_cluster.py`:
- recheck corrects a simulated dropped drive
- recheck does not override a real command issued before it fires


src/zigbee/relay_cluster.h
```diff
@@ -16,6 +17,8 @@ typedef struct {
     relay_t *            relay;
     led_t *              indicator_led;
     uint8_t              indicator_state;
+    hal_task_t           startup_recheck_task;
+    uint8_t              startup_target_on;
 } zigbee_relay_cluster;
```
andrei-lazarov, line 22: Thanks for looking into it!
Not sure if rechecking is the best approach..

Also editing this structure will cause a mismatch when the device receives the update.
It needs an "nvm migration", otherwise it can brick.
But I don't think we ever done those, I suggested to romasku we change how we store the data.. need to work on that

terragady: I was just wanted to play with and pushed only to use CI for building the firmware to not occupy my PC :) I know this is a workaround and not actual fix. Sorry, github suggested the upstream for a PR, didn't mean to bother you yet.

andrei-lazarov: Ah don't worry, play around, just don't brick your devices 🙂

PR #515, opened 2026-09-17
Update device_db.yaml

Hydron-4hv:
lvhy15ix (Aubess 3 gang module) changed from router to end_device for flexibility to choose type in poor signal installations


andrei-lazarov:
Thanks

